PR #98, opened 2015-05-19
Implement class property annotations

ipeychev:


eduardolundgren:
This feature is nice but since it's an old pull request and we never decided anything about it I am closing it. Thanks!

jbalsas:
I was thinking about this the other day... if I remember correctly, we settled on only bringing to the core final JS features and avoid any of those in stage[0-3]. Is that correct?

eduardolundgren:
Yep. It's safer that way.

PR #101, opened 2015-12-04
Improving core.UID_PROPERTY uniqueness.

henvic:
I have made some tests to see how bad it was: https://gist.github.com/henvic/409bd08bf28cba38a862

And it doesn't seem 'that' bad for our needs, but here is a patch anyways.

https://medium.com/@betable/tifu-by-using-math-random-f1c308c4fd9d
8eb7270

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -217,7 +217,7 @@ class core {
  * @type {String}
  * @protected
  */
-core.UID_PROPERTY = 'core_' + ((Math.random() * 1e9) >>> 0);
+core.UID_PROPERTY = 'core_' + (Date.now() % 1e9) + '' + ((Math.random() * 1e9) >>> 0);
 
 /**
  * Counter for unique id.


mairatma:
Nice, makes sense that this is even better. Do you have the results for this same test when run with your change?

@eduardolundgren, can you take a look and confirm that this small change is all right?

henvic:
0 duplicates, as expected.

eduardolundgren:
This should be fine, if it's clear what it does.

@henvic Could you explain why the modulus operation on Date.now() % 1e9?

henvic:
The first digits of Date.now() are always the same during a given decade / year / month / etc, so by taking the modulus we just get rid of most part of the id that doesn't add uniqueness. We lose the side-effect of knowing for certainly when the id was created with easiness, but it really doesn't really matter for any practical reason.

eduardolundgren:
LGTM. Thanks for the change!

mairatma:
Thanks! Pull request merged :)

PR #102, opened 2015-12-04
Making the async module Node compatible.

henvic:


henvic:
Comments added.

yuchi:
I think the correct change would be to use `global.…` instead of `window.…`. In fact `core.isFunction` just uses `typeof`.

Is the build system of Metal polyfilling `global`?

mairatma:
@yuchi, we actually want api.js to work both on browsers and on node, but using `global....` would only work on node, unless we polyfill it like you mentioned. I'd personally prefer changing the check this way to adding a `window.global` variable, though both approaches are fine.

henvic:
@yuchi, you need to use the typeof operator directly here, because if you try to use it inside a function enclosing it (as in core.isFunction) when an undefined variable is passed to the function an exception will be thrown. If we use directly we will get back the undefined value.

yuchi:
@mairatma: I'm not specifically advocating for placing it as `window.global` 😬 both Webpack and Browserify shim it automatically with a local reference.

@henvic that's for sure :)

mairatma:
You're right, we could add it as a local variable for each file as well during the build, but that would make the build more complicated. We prefer leaving the code like this, which would work regardless of how the files are built.

yuchi:
I just wanted to point that while no real standard way to get globals exists, the `global.…` form is way more explicit and… cool! :sunglasses:

henvic:
@mairatma, I have added a more detailed comment.

mairatma:
Thanks @henvic! Just remember to run `gulp lint` next time, there was a small lint error which I've fixed.
I'll see if I can setup github to automatically run this for each pr later, but for now it's manual :)

PR #103, opened 2015-12-22
Revert "Improving core.UID_PROPERTY uniqueness."

henvic:
Fixed on Chrome.

[There's Math.random(), and then there's Math.random()](http://v8project.blogspot.com/2015/12/theres-mathrandom-and-then-theres.html)


eduardolundgren:
Thanks!

PR #148, opened 2016-08-08
Dom.triggerEvent method now reproduces browsers behaviour for the click event to disabled elements - Fixes #134

### LINKED ISSUE #134: Event delegation should skip "disabled" elements
_From @mairatma on July 5, 2016 17:37_

Events don't trigger for disabled elements natively, even when bubbling. Our event delegation doesn't respect that though, since it catches the event in the component's element and doesn't check for disabled states.

Need to investigate which elements are automatically skipped natively (button for sure, but maybe more of them too).

_Copied from original issue:_ https://github.com/metal/metal-events/issues/2

_Copied from original issue: mairatma/metal-lerna#2_

fernandosouza:
Guys. Before merging, please, run `gulp test:browsers`

### DIFF
--- a/packages/metal-dom/src/dom.js
+++ b/packages/metal-dom/src/dom.js
@@ -259,7 +259,7 @@ class dom {
 	 * @param {Element} node Element to be checked.
 	 * @param {string} eventName The event name.
 	 */
-	static isAbleToInteract_(node, eventName) {
+	static isAbleToInteractWith_(node, eventName) {
 		var currElement = node;
 		var isAble = true;
 		var matchesSelector = 'button, input, select, textarea, fieldset';
@@ -709,7 +709,7 @@ class dom {
 	 *   triggered event's payload.
 	 */
 	static triggerEvent(element, eventName, opt_eventObj) {
-		if (dom.isAbleToInteract_(element, eventName)) {
+		if (dom.isAbleToInteractWith_(element, eventName)) {
 			var eventObj = document.createEvent('HTMLEvents');
 			eventObj.initEvent(eventName, true, true);
 			object.mixin(eventObj, opt_eventObj);
--- a/packages/metal-dom/src/dom.js
+++ b/packages/metal-dom/src/dom.js
@@ -254,6 +254,31 @@ class dom {
 	}
 
 	/**
+	 * Verifies if the element is able to trigger the Click event,
+	 * simulating browsers behaviour, avoiding event listeners to be called by dom.triggerEvent method.
+	 * @param {Element} node Element to be checked.
+	 * @param {string} eventName The event name.
+	 */
+	static isAbleToInteract_(node, eventName) {
+		var currElement = node;
+		var isAble = true;
+		var matchesSelector = 'button, input, select, textarea, fieldset';
+
+		if (eventName === 'click') {
+			while (currElement && currElement.tagName !== 'LEGEND') {
+				if (currElement.disabled && dom.match(currElement, matchesSelector)) {
+					isAble = false;
+					break;
+				}
+
+				currElement = currElement.parentNode;
+			}
+		}
+
+		return isAble;
+	}
+
+	/**
 	 * Inserts node in document as last element.
 	 * @param {Element} node Element to remove children from.
 	 */
@@ -684,10 +709,12 @@ class dom {
 	 *   triggered event's payload.
 	 */
 	static triggerEvent(element, eventName, opt_eventObj) {
-		var eventObj = document.createEvent('HTMLEvents');
-		eventObj.initEvent(eventName, true, true);
-		object.mixin(eventObj, opt_eventObj);
-		element.dispatchEvent(eventObj);
+		if (dom.isAbleToInteract_(element, eventName)) {
+			var eventObj = document.createEvent('HTMLEvents');
+			eventObj.initEvent(eventName, true, true);
+			object.mixin(eventObj, opt_eventObj);
+			element.dispatchEvent(eventObj);
+		}
 	}
 
 	/**
--- a/packages/metal-dom/test/dom.js
+++ b/packages/metal-dom/test/dom.js
@@ -625,6 +625,124 @@ describe('dom', function() {
 				dom.triggerEvent(matchedElements[1], 'click');
 				assert.ok(!listener.args[0][0].delegateTarget);
 			});
+
+			it('should not trigger delegate click event for disabled elements', function() {
+				var parent = document.createElement('div');
+				document.body.appendChild(parent);
+
+				var input = document.createElement('input');
+				var select = document.createElement('select');
+				var button = document.createElement('button');
+				input.disabled = true;
+				select.disabled = true;
+				button.disabled = true;
+				parent.appendChild(input);
+				parent.appendChild(select);
+				parent.appendChild(button);
+
+				var listener1 = sinon.stub();
+				var listener2 = sinon.stub();
+				var listener3 = sinon.stub();
+
+				dom.delegate(parent, 'click', input, listener1);
+				dom.delegate(parent, 'click', select, listener2);
+				dom.delegate(parent, 'click', button, listener3);
+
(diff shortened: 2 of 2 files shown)


fernandosouza:
Travis wasn't able to run `gulp test:sauce labs`. Have I broken something? 😢

packages/metal-dom/src/dom.js
```diff
@@ -254,6 +254,31 @@ class dom {
...
+	 * @param {Element} node Element to be checked.
+	 * @param {!String} eventName The event name.
+	 */
+	static isAbleToInteract_(node, eventName) {
+		var currElement = node;
+		var isAble = true;
+		var matchesSelector = 'button, input, select, textarea, fieldset';
+
+		if (eventName === 'click') {
+			while (currElement && dom.match(currElement, matchesSelector) && !dom.match(currElement, 'legend')) {
```
mairatma: It shouldn't be possible for an element to match the `button, input, select, textarea, fieldset` selector and also match `legend`, so this second `match` call can probably be removed, right?

packages/metal-dom/src/dom.js
```diff
@@ -254,6 +254,31 @@ class dom {
 	}
 
 	/**
+	 * Verifies if the element is able to trigger the Click event,
+	 * simualting browsers behaviour, avoiding event listeners to be called by dom.triggerEvent method.
+	 * @param {Element} node Element to be checked.
+	 * @param {!String} eventName The event name.
+	 */
+	static isAbleToInteract_(node, eventName) {
+		var currElement = node;
```
mairatma, line 263: Both `currElement` and `matchesSelector` are only used inside the if clause, so it'd be best to move them there (we don't require var declarations at the top of the function for Metal.js, we instead prefer them to be near their first usage. If you use `let` or `const` instead of  `var` this is even better.

packages/metal-dom/src/dom.js
```diff
@@ -254,6 +254,31 @@ class dom {
 	}
 
 	/**
+	 * Verifies if the element is able to trigger the Click event,
+	 * simualting browsers behaviour, avoiding event listeners to be called by dom.triggerEvent method.
+	 * @param {Element} node Element to be checked.
+	 * @param {!String} eventName The event name.
```
mairatma: Primitive types should be lowercase in docs, and don't require a `!` as they're required by default. So this should be just `{string}`. Just FYI, we're following [google's standard](https://developers.google.com/closure/compiler/docs/js-for-compiler) for docs.

mairatma:
It's great that you manage to find out where this fails, so it's only for `triggerEvent`, right? Makes total sense.

Could you just rename this commit to explain the fix? It's not just reproducing the problem, it's also fixing it, so the name should say that instead. Also the issue number is incorrect (should be #134, not #34). We usually follow the following pattern: `[Explains fix] - Fixes #[issue number]`.

mairatma:
Oh and the travis tests don't work for prs yet, it's a configuration problem with saucelabs, I still need to work on fixing this.

packages/metal-dom/src/dom.js
fernandosouza: In this case I'm testing if a parent of the node passed as argument is a legend tag. Because, according to W3C, if a 'button, input, select or textarea' are children of a disabled 'fieldset' they should be disabled, unless, those children are wrapped by a legend.

packages/metal-dom/src/dom.js
fernandosouza, line 263: Got it! :)

packages/metal-dom/src/dom.js
fernandosouza: Got it! :)

packages/metal-dom/src/dom.js
mairatma: You're checking this on the same element you're checking for `button, input, select, textarea, fieldset` right before though, so if the code reaches `!dom.match(currElement, 'legend')` it means that the tag is already one of `button, input, select, textarea, fieldset`, so it can't be `legend` anymore, right?

packages/metal-dom/src/dom.js
```diff
@@ -254,6 +254,31 @@ class dom {
 	}
 
 	/**
+	 * Verifies if the element is able to trigger the Click event,
+	 * simualting browsers behaviour, avoiding event listeners to be called by dom.triggerEvent method.
```
mairatma: Just noticed a small typo here too: `simulating`.

packages/metal-dom/src/dom.js
mairatma: Also, would be good to have a test for this `legend` case, to make sure it works as expected :)

fernandosouza: The node element will only be a button or input or select or textarea at the first loop interaction. After that, the 'while' will check if currElement is a 'fieldset'. The matchesSelector variable stores the possibles positives match in only one selector. 'Legend' is there just to stop the loop.

Does it make sense? 😕

fernandosouza:
@mairatma. Branch updated.

mairatma:
Just started reviewing :) <br><br>:octocat: _Sent from [GH](http://nodegh.io)._

packages/metal-dom/src/dom.js
```diff
@@ -254,6 +254,31 @@ class dom {
...
+	 * @param {Element} node Element to be checked.
+	 * @param {!String} eventName The event name.
+	 */
+	static isAbleToInteractWith_(node, eventName) {
+		var currElement = node;
+		var isAble = true;
+		var matchesSelector = 'button, input, select, textarea, fieldset';
+
+		if (eventName === 'click') {
+			while (currElement && !dom.match(currElement, 'legend')) {
```
mairatma: Since this is just checking the tag name it would best to do `currElement.tagName !== 'LEGEND'` here instead, since `dom.match` is a heavy operation.

packages/metal-dom/src/dom.js
```diff
@@ -254,6 +254,31 @@ class dom {
...
+				if (dom.match(currElement, matchesSelector) && currElement.disabled) {
```
mairatma: It's best to do the `currElement.disabled` check first, as that's faster than `dom.match` and will fail most times.

packages/metal-dom/test/dom.js
```diff
@@ -625,6 +625,148 @@ describe('dom', function() {
...
+				dom.triggerEvent(input, 'click');
+				dom.triggerEvent(select, 'click');
+				dom.triggerEvent(button, 'click');
+				assert.strictEqual(0, listener1.callCount);
+				assert.strictEqual(0, listener2.callCount);
+				assert.strictEqual(0, listener3.callCount);
+				assert.strictEqual(0, listener4.callCount);
+			});
+
+			it('should trigger delegate click event to an element which its valid parent is disabled unless it is wrapped by an legend element', function() {
```
mairatma: It seems like this description is saying the opposite of what it means. It's saying that the event should trigger unless wrapped by legend, when it's the opposite actually.

That probably happened because the description got confusing. Something like this would work a bit better, for example: `it should trigger delegate click event on elements wrapped by a "legend" tag and with disabled parent`.

packages/metal-dom/src/dom.js
```diff
@@ -254,6 +254,31 @@ class dom {
 	}
 
 	/**
+	 * Verifies if the element is able to trigger the Click event,
+	 * simulating browsers behaviour, avoiding event listeners to be called by dom.triggerEvent method.
+	 * @param {Element} node Element to be checked.
+	 * @param {!String} eventName The event name.
```
mairatma: This should be `{string}` (lowercase for primitive types, and no **!** since primitive types are already non null by default).

fernandosouza:
@mairatma, branch updated. Thx for the tips.

mairatma:
Just started reviewing :) <br><br>:octocat: _Sent from [GH](http://nodegh.io)._

mairatma:
This has been merged. Thanks a lot! :D

PR #166, opened 2016-10-05
Adds Karma Firefox Launcher

fernandosouza:
I think it's useful for detecting FF issues before sending to CI.

### DIFF
--- a/karma-sauce.conf.js
+++ b/karma-sauce.conf.js
@@ -56,7 +56,7 @@ module.exports = function(config) {
 			'packages/metal*/test/**/*.js': ['browserify']
 		},
 
-		browsers: ['Chrome'],
+		browsers: ['Firefox', 'Chrome'],
 
 		browserify: {
 			debug: true,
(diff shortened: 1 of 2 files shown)


mairatma:
Doesn't that work by just running `gulp test:browsers` today? The idea is to have a fast test in `gulp test` on a single browser so you can focus on logic mistakes, and then you can run `gulp test:browsers` to check if there are any browser specific problems. The IE browsers for this command are a bit problematic (probably due to the need of having the right virtual machines installed on Mac) but the others like Firefox and Safari should be working as expected.

fernandosouza:
I did not notice that `gulp test:browsers` could do what I would like. It was useful to me because I could check out only these two browsers because they were the counterpoint of my problem, or, only set FF for a specific debug.

:)

mairatma:
I think it's best to keep this as is then: one command to run in a single browser for fast testing, and another for running in all browsers. Whenever you need to run a specific different browser you can set it up temporarily on your side. It's more work but it's a rarer use case :)

fernandosouza:
Perfect! :)

PR #170, opened 2016-11-09
Gulp format lint from gulp metal

pragmaticivan:


pragmaticivan:
cc @mairatma I was not able to run all the test locally, so I'm opening this temporary PR just to test something here.

mairatma:
Hey @pragmaticivan, I understand that you're saying this the pr is temporary, but just to give you some feedback: 

The change to`object.mixin` looks fine, it needs only more test cases to cover the changes you've made (for example, one that tests that it will throw an error if the first arg is null/undefined, another that tests passing undefined as one of the other arguments, etc).

I prefer not to have the lint/format commit though. The format tool currently has some problems with some of the formatting. It's doing bad things it's not supposed to do, which are the cause of most of the changes in that commit. We need to fix its configuration or switch format tools before allowing all its changes to enter the repo. So we're currently only running `gulp lint` right now.

pragmaticivan:
Thanks @mairatma :D

Issue #168, opened 2016-10-25
Suggestion. Could we have a keyCode map?


fernandosouza:
In order to work easily with keyboard events, it will be nice to have a keyCode map like http://yuilibrary.com/yui/docs/event/key.html

yuchi:
You can steal this as a starting point: https://github.com/smclab/react-faceted-token-input/blob/master/src/key-codes.js

Of course you'd also need the reversed map (`code -> key`).

fernandosouza:
Cool, @yuchi! Thank you! :)

yuchi:
You can access them as a single object with `import * as codes from …`.

fernandosouza:
I began to implement.

mairatma:
Just FYI, Fernando added this as a separate module called [metal-key](https://github.com/metal/metal-key).

PR #175, opened 2016-11-11
Adds special treatment for parameterized keyboard events

fernandosouza:


packages/metal-dom/src/keyboardEvents.js
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * Checks if the given event name is an keyboard event and it has matched key
+ * parameters.
+ * @param {string} eventName The given event name.
+ * @return {boolean} True if eventName matches with an valid keyboard event and
+ * there is at least one parameter.
+ */
+function isParametedKeyboardEvent(eventName) {
+  return /^key(down|up|press):([a-z]+),?([a-z]+)?/g.test(eventName);
```
mairatma: This regex wouldn't work if more than 2 key codes are passed, for example: `'keydown:space,esc,up'`. Something like this would cover one or more key codes, as we need: `/^key(down|up|press):((?:[a-z]+)(?:,[a-z]+)*)$/`.

Also, since this regex is always the same, it'd be good to make it a constant (something like `EVENT_REGEX` that only this file can access) so that it's not recreated every time this function runs.

packages/metal-dom/src/domNamed.js
```diff
@@ -448,6 +449,11 @@ export function on(element, eventName, callback, opt_capture) {
 	if (isString(element)) {
 		return delegate(document, eventName, element, callback);
 	}
+	var keyConfig = getKeyboardEventConfig(eventName);
+	if (keyConfig) {
```
mairatma: This `if` clause has the same logic as the one for `customConfig` below. We could take advantage of this by reusing the `customConfig` variable, like this:

``` js
var customConfig = customEvents[eventName] || getKeyboardEventConfig(eventName);
```

That way we can remove all these extra lines :)

packages/metal-dom/src/keyboardEvents.js
```diff
@@ -0,0 +1,156 @@
+'use strict';
+
+import { object } from 'metal';
+
+export const KEYMAP = {
+  'BACKSPACE': 8,
```
mairatma: These `'` around this object's keys could all be removed, we only add them when strictly necessary.

packages/metal-dom/src/keyboardEvents.js
```diff
@@ -0,0 +1,156 @@
...
+}
+
+/**
+ * Checks if the given event name is an keyboard event and it has matched key
+ * parameters.
+ * @param {string} eventName The given event name.
+ * @return {boolean} True if eventName matches with an valid keyboard event and
+ * there is at least one parameter.
+ */
+function isParametedKeyboardEvent(eventName) {
```
mairatma: `isParameterizedKeyboardEvent`

packages/metal-dom/src/keyboardEvents.js
```diff
@@ -0,0 +1,156 @@
...
+  return KEYMAP[keyName.toUpperCase()];
+}
+
+/**
+ * Breaks down the given parameterized event name and it returns the original
+ * keyboard event name and the matched keys.
+ * @param {string} eventName The given event name.
+ */
+function getEventConfig_(eventName) {
+  let eventData = eventName.split(':');
```
mairatma: Instead of handling the string manually here you could also make use of the regex for this, which would also be safer since it would always follow the pattern indicated there. You can both check if `eventName` matches the regex and get its param values with the same call if you use the regex I've mentioned in the other comment below, and use its `exec` function. For example:

``` js
const matchData = EVENT_REFEX.exec(eventName);
if (matchData) {
  return {
    originalEvent: matchData[1],
    keys: matchData[2].split(',').map(convertKeynameIntoKeyCode_),
    handler: handlerKeys
  };
}
```

packages/metal-dom/src/keyboardEvents.js
```diff
@@ -0,0 +1,156 @@
...
+ * Original listener wrapper. It checks if the pressed key matchs with the given
+ * keys and it trigger the original listener function.
+ * @param {!function(!Object)} callback The original listener that will called
+ * when the event is triggered and if the pressed key and the given keys alias
+ * matches.
+ * @param {event} Normalized event object.
+ * there is at least one parameter.
+ */
+function handlerKeys(callback, event) {
+  if (this.keys.indexOf(event.keyCode) > -1) {
```
mairatma: If we store these keys in an object we'll be able to do faster checks here, avoiding `indexOf` and doing something like this instead: `this.keys[eventKeyCode]`.

packages/metal-dom/test/dom.js
```diff
@@ -359,6 +360,73 @@ describe('dom', function() {
...
 		});
+
+		it('should listen to keyboard events by keyCode alias', function() {
+			var element = document.createElement('input');
+			element.setAttribute('type', 'text');
+			dom.enterDocument(element);
+
+			var listener = sinon.stub();
+			var handle = dom.on(element, 'keydown:enter', listener);
+			dom.triggerEvent(element, 'keydown', {keyCode: KEYMAP['ENTER']});
```
mairatma: Dot notation is prefered whenever possible when using objects, so this should be `KEYMAP.ENTER`.

packages/metal-dom/test/dom.js
```diff
@@ -359,6 +360,73 @@ describe('dom', function() {
 			dom.triggerEvent(element, 'click');
 			assert.strictEqual(0, listener.callCount);
 		});
+
+		it('should listen to keyboard events by keyCode alias', function() {
+			var element = document.createElement('input');
+			element.setAttribute('type', 'text');
```
mairatma: The element you're calling `triggerEvent` on doesn't need to be a text input, so removing this line would make the tests a bit simpler.

packages/metal-dom/test/dom.js
```diff
@@ -359,6 +360,73 @@ describe('dom', function() {
...
+			dom.enterDocument(element);
+
+			var listener = sinon.stub();
+			var handle = dom.on(element, 'keyup:play', listener);
+			dom.triggerEvent(element, 'keyup', {keyCode: KEYMAP['SPACE']});
+			assert.strictEqual(0, listener.callCount);
+		});
+
+		it('should not trigger the listener to an inexistent key alias', function() {
```
mairatma: This test seems to be the same as the one in line 387, I think you forgot to remove it :)

mairatma (commented):
Nice, thanks @fernandosouza, this is working exactly as it should! I think it will be very helpful.

packages/metal-dom/src/keyboardEvents.js
mairatma: Oh, and btw, you may already know this but, in case you don't, there are some really helpful sites that you can use to test your regexes. I always use this one when checking if mine are covering what I need: https://regex101.com/. They not only check if some string matches the regex, but also shows what params would be extracted from them, explaining how that's done by the regex algorithm step by step.

packages/metal-dom/src/keyboardEvents.js
mairatma: Oh and one more thing I've just thought about, we need to export this KEYMAP from the `all/dom.js` file, otherwise people won't be able to access it. I don't think we should export the other functions available here, this object is the one that would be helpful for others.

mairatma:
Ah, and could you link the github [issue](https://github.com/metal/metal.js/issues/168) with this commit? The issue mentions the key map specifically, which this already covers.

packages/metal-dom/src/keyboardEvents.js
fernandosouza: Sure, @mairatma. Done.

packages/metal-dom/test/dom.js
fernandosouza: Yes. Thx

packages/metal-dom/test/dom.js
fernandosouza: Done

packages/metal-dom/src/keyboardEvents.js
yuchi: I have a small note on this. if instead of using an object we exported keymaps one by one we could have rollup automatically dead-code-eliminate all the stuff we don‘t use. It’s just a few bytes but it’s still something.

yuchi: Sorry @mairatma and @fernandosouza, I have to take back what I said. Since we’re talking about `keyup:KEY` events, you need to have the whole thing at runtime. Rollup couldn’t help here.

This doesn’t mean that having the source exporting stuff like that is good anyway. You can later `import * as keys from '../keys'` to access them as a single object.

mairatma: Yes, I think Rollup won't help since we use the map to check if the key is valid in the code anyway, so we'll always import the whole object. But, like you said, exporting them separately is still a good option, we can use this kind of import statement when we need an object with all of them.

packages/metal-dom/src/keyboardEvents.js
fernandosouza: Thx for the spelling correction, @mairatma. But this method was removed due to the changing of the implementation.

packages/metal-dom/src/keyboardEvents.js
fernandosouza: Thx for the tip, @mairatma. Done.

packages/metal-dom/src/domNamed.js
fernandosouza: Thx. Working on it.

packages/metal-dom/src/keyboardEvents.js
fernandosouza: Yeah, you both are right, guys. I did not remember that we can achieve this behavior using `import`. Thx.

packages/metal-dom/src/keyboardEvents.js
fernandosouza: Done.

packages/metal-dom/src/domNamed.js
```diff
@@ -296,6 +297,28 @@ export function exitDocument(node) {
...
+ * Search by a custom event already registered and returns its configuration object.
+ * @param {!string} eventName The custom event name used to find the configuration
+ * object as an object key or by executing a regular expression.
+ * @return {Object} The custom event configuration object or undefined if the given
+ * eventName doesn't match with any custom event name.
+ */
+function getCustomEventData(eventName) {
+	var customConfig = customEvents[eventName];
+	if (!customConfig) {
+		for (var i = customUnamedEvents.length; i--; ) {
```
mairatma: Any specific reason to go through these events in reverse order? I think it'd make more sense to loop them in the order that they were registered, so that the precedence is clear in that sense.

packages/metal-dom/src/domNamed.js
```diff
@@ -488,12 +511,20 @@ export function parent(element, selector) {
 
 /**
  * Registers a custom event.
- * @param {string} eventName The name of the custom event.
+ * @param {!string|!Object} eventName The name of the custom event or a regular
```
mairatma: We could be more specific in the type here, like this: `{!RegExp|string}`.

packages/metal-dom/src/domNamed.js
```diff
@@ -488,12 +511,20 @@ export function parent(element, selector) {
...
+ * expression object.
  * @param {!Object} customConfig An object with information about how the event
```
mairatma: This can now also be a function, so the param type and description should be updated accordingly.

packages/metal-dom/src/events.js
```diff
@@ -2,6 +2,54 @@
...
+}
+
+/**
+ * Creates an event configuration object to deail with keyboard events.
+ * @param {!Array} MatchedRegexInfo The information extracted from the regex
+ * that has matched with the parameterized event name.
+ * @param {object} The custom keyboard event configuration that has all the
+ * necessary informations for the handler function.
+ */
+function createCustomKeyboardEventConfig(MatchedRegexInfo) {
```
mairatma: Uppercase variables are usually only used for global constants or constructor functions. This is just an array, so the name should be `matchedRegexInfo` instead.

packages/metal-dom/src/events.js
```diff
@@ -2,6 +2,54 @@
...
+  });
+
+  return keys;
+}
+
+/**
+ * Creates an event configuration object to deail with keyboard events.
+ * @param {!Array} MatchedRegexInfo The information extracted from the regex
+ * that has matched with the parameterized event name.
+ * @param {object} The custom keyboard event configuration that has all the
```
mairatma: I think this was supposed to be the `@return` doc, not `@param` :)

packages/metal-dom/src/events.js
```diff
@@ -2,6 +2,54 @@
...
+const REGEX_EVENT = /^(key+(down|up|press)):((,?[a-z])+)/;
+
+/**
+ * Returns the relative keyCode from each key name in an array.
+ * @param {Array<string>} keyNames An array with the key names.
+ */
+function convertKeynamesToKeyCode_(keyNames) {
+  let keys = {};
+
+  keyNames.forEach((value, index) => {
```
mairatma: It's best to avoid using `forEach` in places that may be called multiple times by the user, to avoid allocating temporary anonymous functions like this one.

packages/metal-dom/src/events.js
```diff
@@ -2,6 +2,54 @@
...
+/**
+ * Original listener wrapper. It checks if the pressed key matchs with the given
+ * keys and it trigger the original listener function.
+ * @param {!function(!Object)} callback The original listener that will called
+ * when the event is triggered and if the pressed key and the given keys alias
+ * matches.
+ * @param {event} Event object.
+ */
+function handlerKeys(callback, event) {
+  if (this.keys[event.keyCode]) {
```
mairatma: We always set the `customType` property of the event before triggering it, in case the developer wants to know  it. It'd be good to do this here as well, same way we do for the other custom events on this file: `event.customType = eventName`. Note that `eventName` should be the name that was passed to `dom.on` or `dom.delegate`, for example: `keydown:enter,esc`.

packages/metal-dom/test/events.js
```diff
@@ -102,4 +109,86 @@ describe('Custom Events', function() {
...
+
+		var listener = sinon.stub();
+		var handle = dom.on(element, 'keydown:enter', listener);
+		dom.triggerEvent(element, 'keydown', {keyCode: KEYMAP.ENTER});
+		assert.strictEqual(1, listener.callCount);
+	});
+
+	it('should delegate parameterized keyboard event', function() {
+		var listener = sinon.stub();
+		var handle = dom.delegate(this.element4, 'keydown:enter', this.element5, listener);
```
mairatma: Couldn't this test use `element2` and `element3` instead of creating two new elements? That would simplify things.

packages/metal-dom/test/events.js
```diff
@@ -102,4 +109,86 @@ describe('Custom Events', function() {
...
+		var element = document.createElement('input');
+		dom.enterDocument(element);
+
+		var listener = sinon.stub();
+		var handle = dom.on(element, 'keyup:play', listener);
+		dom.triggerEvent(element, 'keyup', {keyCode: KEYMAP.SPACE});
+		assert.strictEqual(0, listener.callCount);
+	});
+
+	it('should not trigger the listener to an inexistent key alias', function() {
```
mairatma: The test description seems wrong here, it's the same description as the one in the previous test, but it's actually testing that it won't trigger a listener for the wrong key alias, not a nonexistent one.

packages/metal-dom/src/domNamed.js
mairatma: I think it could be since they're regexes, so different custom events could actually match the same event name. From profiling Metal.js I can see that looping arrays haven't made a difference to performance, and this array is going to be usually very small (almost always with only one element for now even, unless a developer adds his own custom event), so going backwards shouldn't make a difference.

packages/metal-dom/src/events.js
fernandosouza: Done

packages/metal-dom/src/events.js
fernandosouza: Done

packages/metal-dom/src/events.js
fernandosouza: Done

packages/metal-dom/src/events.js
fernandosouza: Done

packages/metal-dom/test/events.js
fernandosouza: Sure.

packages/metal-dom/test/events.js
fernandosouza: Done

packages/metal-dom/test/events.js
fernandosouza: I'm going to use element1 and inside the test I will append the input. Ok?

mairatma:

packages/metal-dom/src/events.js
```diff
@@ -2,6 +2,45 @@
 
 import { registerCustomEvent } from './dom';
 import features from './features';
+import * as KEYMAP from './keyConstants';
+
+const REGEX_EVENT = /^(key+(down|up|press)):((,?[a-z])+)/;
+
+/**
+ * Returns the relative keyCode from each key name in an array.
+ * @param {Array<string>} keyNames An array with the key names.
```
mairatma, line 11: This is missing the `@return` doc. Also, the description says that this will return the result `in an array`, but it's actually `as an object`.

mairatma (commented):
@fernandosouza I was about to merge this, it looks good, but I then realized that `gulp lint` is complaining of a few errors. Can you fix those together with the single inline comment I've made?

packages/metal-dom/src/events.js
fernandosouza: Ops!!! Sure, @mairatma.

eduardolundgren (commented):
Previously, a switch-case statement could only have up to 128 case-clauses, more than that and the function containing the switch statement was not optimizable. This one is lot less than that, we are fine here.

eduardolundgren:
I was talking to @fernandosouza about this feature. I am not sure if this should be part of the core.
This seems will cause more problems than benefits being on the core.
wdyt?

mairatma:
Yes, you may be right @eduardolundgren. What do you think about just leaving the logic that allows custom events with regexes to be registered in the core, but leaving the actual key event logic outside as a separate optional module you can import? That way this event will be available, but not by default.

As for the regexes, I think that Fernando has already explained to you, but they're required because the event name may be of the form `key:enter,esc` for example, so it's dynamic.

yuchi:
I personally enjoyed key-namespaced events in YUI, but they had little to nothing documentation so they were never to be found in the wild.

eduardolundgren:
@mairatma I would rather keep this as a external module, also I don't think we even need the regex support on the core to allow the external module that adds this functionality, this new module can register `N` number of events dynamically during load instead of trying to infer based on regex rules.

eduardolundgren:
As additional information the external module once loaded will perform:

```js
for (...N) {
  registerCustomEvent('key:' + keyCodeLabel);
}
```

It will be simpler to do this logic for building the event names on the module side instead of regexes.

wdyt?

mairatma:
Just FYI, it was decided that this will be done as a [separate module](https://github.com/metal/metal-key) instead, so closing the pull request.

Issue #178, opened 2016-12-13
Feature request: Add a `willRender` method to (Soy)Components


yuchi:
In JSX Components there’s a simple moment to compute values to be used in the actual rendering logic:

```js
class MyButton extends JSXComponent {
  render() {
    const className = this.props.primary ? 'btn btn-primary' : 'btn';
    return <button class={ className }>Hallo</button>;
  }
}
```

This is completely missing in Soy components, and that means that components’ properties more or less need to be the ones you use in the Soy template.
Since in Soy is not possible to call functions to manipulate the data, this makes difficult to have properties that depend on the value of another one.

Examples include:

- filtered, split or manipulated arrays;
- transform a file size (in bytes) in a human readable string (“23GB”)

- - -

The proposal is to add a new `willRender` method (similar to React’s [`componentWillUpdate`](https://facebook.github.io/react/docs/react-component.html#componentwillupdate)) that can be used to update the properties passed down to the template.

```soy
{template .render}
  <span>The file is {$humanReadableFileSize} big</span>
{/template}
```
```js
class FileInfo extends Component {
  willRender() {
    return {
      "humanReadableFileSize": humanize(this.fileSize)
    };
  }
}
FileInfo.STATE = {
  "fileSize": {
    value: 0
  }
};
Soy.register(FileInfo, templates);
```

mthadley:
This would be great to have, and I see a real need for this as well. 

I've seen `sync` methods used to try and emulate this behavior in the past, but it is certainly not as efficient, since it means that the component will render at least twice. Once for the first property change, and then again for the derived property after the `sync` method was called.

yuchi:
Exactly. The price to pay if obviously a harder process to keep browser and server code in sync

yuchi:
I’m implementing it, but I took a different approach.

#### Mutability

First I don’t like having a method that can mutate the object that will be sent to the template. I’d prefer the method to act as if data would be immutable.
We have three options:

1. the method gets passed the real template data object, and it can (should) mutate;
2. the method gets passed the real template data object, but must return what becomes the actual data passed to the template — it’s up to the author to choose to either
    - create a new object mixing in the original one with custom keys
    - mutate the original one or
    - just pass new data;
3. the method gets passed the real (or a clone) template data object and what the method returns is then mixed with the template data and passed to the template.

I personally prefer (and implemented) option N. 2.

#### Method Name

[...]

PR #179, opened 2016-12-13
Implement "getTemplateData"

yuchi:
See #178 for reference.

### DIFF
--- a/packages/metal-soy/src/Soy.js
+++ b/packages/metal-soy/src/Soy.js
@@ -62,7 +62,7 @@ class Soy extends IncrementalDomRenderer.constructor {
 			}
 		}
 
-		if (component.getTemplateData) {
+		if (isFunction(component.getTemplateData)) {
 			data = component.getTemplateData(data) || data;
 		}
 
--- a/packages/metal-soy/src/Soy.js
+++ b/packages/metal-soy/src/Soy.js
@@ -40,6 +40,8 @@ class Soy extends IncrementalDomRenderer.constructor {
 	 * template call's data. The copying needs to be done because, if the component
 	 * itself is passed directly, some problems occur when soy tries to merge it
 	 * with other data, due to property getters and setters. This is safer.
+	 * Also calls the component's "getTemplateData" is called to let it change the
+	 * data passed to the template.
 	 * @param {!Component} component
 	 * @param {!Array<string>} params The params used by this template.
 	 * @return {!Object}
@@ -59,6 +61,11 @@ class Soy extends IncrementalDomRenderer.constructor {
 				data[params[i]] = component[params[i]].bind(component);
 			}
 		}
+
+		if (component.getTemplateData) {
+			data = component.getTemplateData(data) || data;
+		}
+
 		return data;
 	}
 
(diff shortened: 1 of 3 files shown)


packages/metal-soy/src/Soy.js
```diff
@@ -40,6 +40,8 @@ class Soy extends IncrementalDomRenderer.constructor {
 	 * template call's data. The copying needs to be done because, if the component
 	 * itself is passed directly, some problems occur when soy tries to merge it
 	 * with other data, due to property getters and setters. This is safer.
+	 * Also calls the component's "getTemplateData" is called to let it change the
```
mairatma: The wording here is a bit weird, maybe just remove teh `is called` part? :)

mairatma (commented):
This pull request is very good, it's a very small change that can already be of much help. Thanks a lot for working on this :)

I just have a single inline comment, but @eduardolundgren, can you take a look as well and say what you think?

packages/metal-soy/src/Soy.js
yuchi: Sorry, wrote it while listening to the training so I probably wrote it in two phases…

Gosh I made a conflict merge artifact without using git…

yuchi:
Thank you @mairatma, but please have a look at the discussion in #178 before accepting this.

packages/metal-soy/src/Soy.js
yuchi: Should be fixed now

packages/metal-soy/src/Soy.js
```diff
@@ -59,6 +61,11 @@ class Soy extends IncrementalDomRenderer.constructor {
 				data[params[i]] = component[params[i]].bind(component);
 			}
 		}
+
+		if (component.getTemplateData) {
```
yuchi: @mairatma should `isFn` be used here? Is usually in Metal the presence of a property defined by the contract enough?

mairatma: `isFunction` would indeed be safer. We usually just check for the presence, but I think it'd be best to use `isFunction` everywhere actually.

eduardolundgren:
LGTM. This feature is nice to have, just keep in mind that when template data is mutated on the client side it would require to reflect the server render state, which is okay.

For the method name I kinda liked the `willRender` but since we have on Component `setState`, `getState`, maybe this might reference the name `State`? Few extra options:

* `beforeRender(data)`
* `formatState(data)`
* `mutateState(data)`
* `prepareState(data)`
* `prepareStateForRender(data)`
* `willRender(data)`

wdyt?

yuchi:
I find the `prepare*` prefix way more apt. I vote for `prepareStateForRender`.

brunobasto:
I like prepareSate better I think.

brunobasto:
On a second thought, maybe formatState makes more sense.

eduardolundgren:
Also, is this just for Soy? Would make sense to have it on the lifecycle for JSX as well?

yuchi:
@eduardolundgren please have a look at #178 where this topics are discussed. It is exclusively for Soy because in other contexts it would make no sense, or even worse it would be a bad practice.

eduardolundgren:
@yuchi I've missed that paragraph on your comment. Make sense. Now I guess it's only the naming left, then we can merge.

packages/metal-soy/src/Soy.js
eduardolundgren: Yep, `isFunction` is needed here.

yuchi: I’ll fix this as a reminder, I’m gonna close this PR

yuchi:
I’m gonna close this, waiting for the end of the poll in #178.

Don‘t forget to cast your vote!

PR #180, opened 2016-12-14
Refactor es6

bryceosterhaus:
Refactored vars to use es6 and also changed string concatenations to template literals.

Let me know if these changes are okay, I also ran `gulp format` and seemed to get a lot of strange changes, let me know if I should remove that.

### DIFF
--- a/packages/metal-component/src/events/events.js
+++ b/packages/metal-component/src/events/events.js
@@ -58,9 +58,9 @@ export function getComponentFn(component, fnName) {
 	if (isFunction(component[fnName])) {
 		return component[fnName].bind(component);
 	} else {
-		console.error('No function named "' + fnName + '" was found in the ' +
-			'component "' + getFunctionName(component.constructor) + '". Make ' +
-			'sure that you specify valid function names when adding inline listeners.'
+		console.error(`No function named ${fnName} was founc in the component
+			"${getFunctionName(component.constructor)}". Make sure that you specify
+			valif function names when adding inline listeners`
 		);
 	}
 }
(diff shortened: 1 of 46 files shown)


yuchi:
I ❤️ this `const`-by-default thing.

eduardolundgren:
LGTM, did you run `gulp format lint`? Thanks!

bryceosterhaus:
@eduardolundgren yeah the second commit is the result from running `gulp format lint` but it seemed to produce a strange output.

packages/metal-component/src/events/events.js
```diff
@@ -58,9 +58,9 @@ export function getComponentFn(component, fnName) {
 	if (isFunction(component[fnName])) {
 		return component[fnName].bind(component);
 	} else {
-		console.error('No function named "' + fnName + '" was found in the ' +
-			'component "' + getFunctionName(component.constructor) + '". Make ' +
-			'sure that you specify valid function names when adding inline listeners.'
+		console.error(`No function named ${fnName} was founc in the component
```
mairatma, line 61: There were a couple of minor errors here when converting to the new format:
* `found` became `founc`
* `valid` became `validf`

These are the only ones I found though, so I can just fix it myself and merge everything now :)

mairatma:

mairatma (commented):
Yes, I've stopped using `gulp format` for a while because of the strange output, waiting for when I would have time to investigate why this was happening and fix it. It's been a long time already, and in the end we're now missing some real fixes from `gulp format` by not using it, so I think it's best to just go back to using it, even if it's weird. Once it's fixed we'll do another format change. So let's keep these changes for now.

mairatma:
This has been [merged](https://github.com/metal/metal.js/commits/master), thanks!

PR #181, opened 2016-12-16
Implement "prepareStateForRender" — Fixes #178

### LINKED ISSUE #178: Feature request: Add a `willRender` method to (Soy)Components

yuchi:
Let Soy components define "prepareStateForRender" to manipulate data:

While in JSXComponents there’s a place to compute derived values from the state just before rendering (that’s the "render" method), on Soy components there’s no way to do the same.

This change makes the Soy renderer to call a "prepareStateForRender" method on the component, passing the data that would be used for the template and trusting the implementation to do one of the following:
- mutate the original template data and return it;
- mutate the original template data and return undefined;
- create a new object and return it (eventually mixing in the original data).

Fixes #178

### DIFF
--- a/packages/metal-soy/src/Soy.js
+++ b/packages/metal-soy/src/Soy.js
@@ -54,6 +54,7 @@ class Soy extends IncrementalDomRenderer.constructor {
 			}
 			data[key] = value;
 		});
+
 		for (let i = 0; i < params.length; i++) {
 			if (!data[params[i]] && isFunction(component[params[i]])) {
 				data[params[i]] = component[params[i]].bind(component);
--- a/packages/metal-soy/src/Soy.js
+++ b/packages/metal-soy/src/Soy.js
@@ -40,6 +40,8 @@ class Soy extends IncrementalDomRenderer.constructor {
 	 * template call's data. The copying needs to be done because, if the component
 	 * itself is passed directly, some problems occur when soy tries to merge it
 	 * with other data, due to property getters and setters. This is safer.
+	 * Also calls the component's "prepareStateForRender" to let it change the
+	 * data passed to the template.
 	 * @param {!Component} component
 	 * @param {!Array<string>} params The params used by this template.
 	 * @return {!Object}
@@ -60,7 +62,12 @@ class Soy extends IncrementalDomRenderer.constructor {
 				data[params[i]] = component[params[i]].bind(component);
 			}
 		}
-		return data;
+
+		if (isFunction(component.prepareStateForRender)) {
+			return component.prepareStateForRender(data) || data;
+		} else {
+			return data;
+		}
 	}
 
 	/**
--- a/packages/metal-soy/test/Soy.js
+++ b/packages/metal-soy/test/Soy.js
@@ -11,6 +11,7 @@ import { Events as EventsComponent } from './assets/Events.soy.js';
 // on it.
 // TODO: We should have a better dependency management for soy files so that
 // the order in which they're required doesn't matter.
+import { ComputedData as ComputedDataComponent } from './assets/ComputedData.soy.js';
 import { ExternalTemplate as ExternalTemplateComponent } from './assets/ExternalTemplate.soy.js';
 import { Functions as FunctionsComponent } from './assets/Functions.soy.js';
 import { HtmlContent as HtmlContentComponent } from './assets/HtmlContent.soy.js';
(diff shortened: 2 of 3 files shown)


packages/metal-soy/test/Soy.js
```diff
@@ -111,6 +112,27 @@ describe('Soy', function() {
 					done();
 				});
 			});
+
+			it('should pass state values to "getTemplateData" and use them in the template', function(done) {
```
mairatma: This still uses the old `getTemplateData` name :)

mairatma (commented):
Thanks a lot @yuchi, this will be very useful :D

I've only added one inline comment, and besides that can you just rename your commits so that they follow the format we use? We basically write commit names as if they've started with the word `If`, so in your case they'd be something like: `Applies source formatting`, `Lets ...` and `Add tests for "prepareStateForRender"`.

yuchi:
Sorry I'm used to `if this commit is applied it will…`.

packages/metal-soy/test/Soy.js
yuchi: Oops!

packages/metal-soy/src/Soy.js
```diff
@@ -60,7 +62,12 @@ class Soy extends IncrementalDomRenderer.constructor {
 				data[params[i]] = component[params[i]].bind(component);
 			}
 		}
-		return data;
+
+		if (isFunction(component.prepareStateForRender)) {
```
yuchi, line 66: @mairatma Didn’t want to make `data` mutable, so I’m branching the return here. Tell me if this is ok with you.

packages/metal-soy/test/Soy.js
```diff
@@ -11,6 +11,7 @@ import { Events as EventsComponent } from './assets/Events.soy.js';
 // on it.
 // TODO: We should have a better dependency management for soy files so that
 // the order in which they're required doesn't matter.
+import { ComputedData as ComputedDataComponent } from './assets/ComputedData.soy.js';
```
yuchi, line 14: Could you make sure this is the right line to place the `import`? I simply sorted them alphabetically.

packages/metal-soy/test/Soy.js
yuchi: Fixed.

packages/metal-soy/test/Soy.js
mairatma: Yes, that's correct, I've just realized that this TODO isn't necessary anymore, all imports can already be in alphabetical order today :)

packages/metal-soy/src/Soy.js
mairatma: Sure, that's fine.

mairatma:

mairatma (approved):
Thanks a lot, merging now!

fernandosouza:
Hey, guys. What do you think to add a documentation for this method in MetalJS website? I can do it.

cc. @yuchi, @jbalsas.

yuchi:
If you could it would be great!

fernandosouza:
Of course, I can. For me, it seems a Lifecycle method. Isn't it?

yuchi:
It
- is a lifecycle method specific for Soy components
- enables computed data without wiring events and `sync*` stuff
- doesn't automatically renders on changes on computed properties so you still need to use the source properties in the template to have changes reflected in the UI (PR to solve that coming)

fernandosouza:
Thank you, @yuchi. I'm going to write it. 😃

PR #182, opened 2016-12-16
Switch to using eslint

bryceosterhaus:
Dependent on https://github.com/metal/gulp-metal/pull/25


packages/metal-state/test/validators.js
```diff
@@ -211,10 +211,12 @@ describe('validators', function() {
 		const NAME = 'name';
 		const PARENT_COMPONENT_NAME = 'parentComponent';
 
+		/* eslint-disable */
 		const ERROR_MESSAGE = `Error: Warning: Invalid state passed to '${NAME}'. ` +
```
mairatma, line 214: Just curious, why does eslint complain about this? Is it an anti pattern to use the new string templates with concatenation?

bryceosterhaus:
Just updated to include `{ "allowTemplateLiterals": true }` for strings

gulpfile.js
```diff
@@ -21,8 +22,7 @@ metal.registerTasks({
 	bundleFileName: 'metal.js',
 	formatGlobs: codeGlobs,
 	karma: require('karma'),
-	// TODO: Find a way to lint jsx files (maybe use eslint instead of jshint).
-	lintGlobs: codeGlobs.concat('!packages/metal-jsx/test/**/*.js'),
```
mairatma, line 25: I'm so glad we can remove that now :)

mairatma (commented):
I just thought of something that might be useful while we're still in this transition phase between jshint and eslint. What if we add an option to **gulp-metal**, something like `useEslint`, that when set to `true` will have `gulp lint` use eslint instead of jshint? That way we don't need to know which task to use for each repo, we can just set this for all that were converted to eslint and keep using `gulp lint` on all of them. Could you do that change on **gulp-metal** as well, @bryceosterhaus? Once that's in we could publish it and merge this too.

bryceosterhaus:
@mairatma  updated pr.

Also see https://github.com/metal/gulp-metal/pull/28

mairatma:
Perfect, let's just wait for **gulp-metal** to be released with that change. I'll do it today, after the complexity [pr](https://github.com/metal/gulp-metal/pull/27) is in as well.

mairatma:
@bryceosterhaus, I was about to merge this but then I realized that the commit names are not in the expected format yet. Could you rename them to that they follow the usual pattern (that is, assume the phrase starts with `It`, as in `Adds eslintrc and remove jshintrc`)?

bryceosterhaus:
@mairatma just pushed changes to commit names. Let me know if those are better.

Thanks!

mairatma:

mairatma:
Thanks, pr merged :)

PR #187, opened 2016-12-22
Upgrades incremental dom to latest stable version

pragmaticivan:


pragmaticivan:
This version uses `this` instead `window`. 

Also not clear why we're not using the node module instead copying the file source.

`const IncrementalDOM = require('incremental-dom')`;

packages/metal-incremental-dom/src/incremental-dom.js
```diff
@@ -20,8 +18,10 @@ import { globals } from 'metal';
  */
 
 (function (global, factory) {
-  (factory((global.IncrementalDOM = global.IncrementalDOM || {})));
-}(globals.window, function (exports) { 'use strict';
+  typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
```
mairatma, line 21: These lines will probably cause some problems with some build types we support (I believe that for gulp-metal's globals build specifically, or maybe AMD as well). That's why I used to change it manually to the lines you've replaced here:

```js
(function (global, factory) {
   (factory((global.IncrementalDOM = global.IncrementalDOM || {})));
 }(globals.window, function (exports) { 'use strict';
```

packages/metal-incremental-dom/src/incremental-dom.js
```diff
@@ -666,9 +666,7 @@ import { globals } from 'metal';
    * @param {?Object<string, !Element>} keyMap
    */
   var removeChild = function (node, child, keyMap) {
-    if (child.parentNode === node) {
```
mairatma, line 669: This was actually an important fix that they did in incremental-dom. Without it some places throw errors. It's just making sure that `removeChild` won't be called if the given node is not the parent. I don't think we should remove that.

mairatma:
We're not using the node module because it took a long time to be updated, and because the format it exports causes problems with some of our build methods (as I explained inline, specifically because of using `this` instead of `window`).

packages/metal-incremental-dom/src/incremental-dom.js
pragmaticivan: Just saw in your latest commit, in the repo, looks like they haven't published in the CDN yet.

packages/metal-incremental-dom/src/incremental-dom.js
pragmaticivan: @mairatma could you explain why use `window` instead this in this case?

pragmaticivan: Ahh got it in your last comment. Will close this PR. Thanks.

pragmaticivan:
@mairatma I'm trying to use that dependency with nodejs for some tests, but

PR #191, opened 2017-01-04
Updates Android 4.4 emulator

fernandosouza:

PR #192, opened 2017-01-04
Reverting commit

fernandosouza:


pragmaticivan:
@fernandosouza why are you reverting everything? the only part related to the last PR was the Android emulator.

fernandosouza:
Hey, @pragmaticivan. The only new commit is the last one. I think those 3 previous commits shown in the history is related to my origin/master branch that shouldn't be updated with the upstream/master.

I will check it out. Thx

fernandosouza:
Hey @eduardolundgren, with this pull tests have passed https://travis-ci.org/metal/metal.js.

What would be a good message for this commit? Could be something like: `Reverts commit 'Older commit message'`

pragmaticivan:
maybe `Reverts updates on Android 4.4 emulator` ?

fernandosouza:
That's a good one too, @pragmaticivan. My idea is to have a clear connection with the commit we are reverting.

Thx.

pragmaticivan:
Also sould be godd adding the merged PR reference in the description.

fernandosouza:
You mean: `Reverts pr #191` or `Reverts updates on Android 4.4 emulator #191`.

fernandosouza:
The tests are falling again and they are taking more time than expected, only after rewrite the commit message. I'm still trying to figure out what is happening.

pragmaticivan:
I think it's a problem with saucelabs infrastructure. I also received an email from the Engineering manager about some maintenance.

fernandosouza:
Yes, @pragmaticivan. I think that's it. But I contacted the support team and I'm doing what they say to see if it fix the problem. Let's see.

fernandosouza:
Hey @eduardolundgren and @pragmaticivan. I think we finally got the problem. https://travis-ci.org/metal/metal.js

I added `-B` (--no-ssl-bump-domains) flag on Saucelabs connection.

Tests weren't running in Android platforms because of this:
screen shot 2017-01-04 at 19 48 42

Here is a link to a similar issue: https://support.saucelabs.com/customer/portal/questions/14577625-hide-the-safari-self-signed-certificate-warning

.travis.yml
```diff
@@ -2,7 +2,9 @@ language: java
 jdk:
   - oraclejdk8
 addons:
-  sauce_connect: true
+  sauce_connect:
+    username: alloyui
```
eduardolundgren: Do we need to specify the username for the fix?

fernandosouza: Here is what Travis specify for this behaviour: https://docs.travis-ci.com/user/sauce-connect/

mairatma: @fernandosouza no, the tests for ios were just not running for some reason, so maybe it's because of some security dialog like this. Maybe this `no_ssl_bump_domains` fix will also work for ios, so it's worth trying it out after this I guess.

fernandosouza: Until now we only have this problem on Android platforms, but there is a similar issue related to iOS in another project.

https://support.saucelabs.com/customer/portal/questions/14577625-hide-the-safari-self-signed-certificate-warning

eduardolundgren: @fernandosouza I was just curious why you added the username if the issue was not related to that. Could you check if it works without the username, to keep it cleaner? Thanks.

fernandosouza: Sure.

fernandosouza: @eduardolundgren, everything looks fine now. 

https://travis-ci.org/metal/metal.js

PR #193, opened 2017-01-05
wip - Adds unscape feature

fernandosouza:


brunobasto:
@fernandosouza, I noticed the commit message says "wip". Are you still working on it?

jbalsas:
Hey @brunobasto, if I'm not mistaken, this was being done to solve https://github.com/liferay/senna.js/issues/185. Since that got fixed in a different way, this should no longer be needed, but it can still be useful at some point.

I guess we should keep this out of `metal.js` until we see the real need for it, but what do you guys think?

brunobasto:
Hey @jbalsas, I agree to keep this out of the core. This could even be a separate repo.

fernandosouza:
Yeah, @jbalsas, you are right. I was working on it with Eduardo. We solved the problem in Senna in a different way like you said. 

I was intending to close this pull, but if you think it will be helpful at some point we can keep it.

jbalsas:
Hey @fernandosouza, thanks for the clarification. Let's close it for now and see if we ever need it again in the future 😉

fernandosouza:
Ok, sounds good. :)

PR #194, opened 2017-01-06
Adds hook for devtools extension

bryceosterhaus:
Let me know if you can foresee any issues with these changes. 

Devtool extension is still a work in progress, but having this in Metal will help speed up development of the extension.

See https://github.com/bryceosterhaus/metal-devtools for reference


eduardolundgren:
LGTM. It shouldn't cause any issue since it's just a optional hook. Let's see if @mairatma has any input on that, otherwise we can merge and release a patch with it.

eduardolundgren:
I am going a head and merge it and if @mairatma has any input we can fix it.

PR #196, opened 2017-01-11
Updates for latest metal-devtools changes

bryceosterhaus:
Changing the API slightly for metal-devtools to work better with components being added while the page has already been rendered. Once merged, I will release a new version of devtools.

PR #197, opened 2017-01-13
Updates for latest metal-devtools changes

bryceosterhaus:
See https://github.com/metal/metal.js/pull/196

Using method on `window.__METAL_DEV_TOOLS_HOOK__` object instead so that we can keep track of all root components.

@jbalsas , I decided to not add the backwards compatibility to the original release of metal-devtools because. That release had some major bugs and it isn't expected that anyone would use that.

### DIFF
--- a/packages/metal-component/src/Component.js
+++ b/packages/metal-component/src/Component.js
@@ -439,8 +439,8 @@ class Component extends EventEmitter {
 		}
 		const instance = new Ctor(config, false);
 
-		if (window.__METAL_DEV_TOOLS_HOOK__) {
-			window.__METAL_DEV_TOOLS_HOOK__(instance);
+		if (window.__METAL_DEV_TOOLS_HOOK__ && window.__METAL_DEV_TOOLS_HOOK__.addRoot) {
+			window.__METAL_DEV_TOOLS_HOOK__.addRoot(instance);
 		}
 
 		instance.renderComponent(element);
--- a/packages/metal-component/test/Component.js
+++ b/packages/metal-component/test/Component.js
@@ -940,9 +940,12 @@ describe('Component', function() {
 		assert.ok(!Component.isComponentCtor(fn.bind(this)));
 	});
 
-	it('should pass instance of component to __METAL_DEV_TOOLS_HOOK__  in Component.render', function() {
-		var hookStub = sinon.stub();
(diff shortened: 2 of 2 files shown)


packages/metal-component/src/Component.js
```diff
@@ -439,8 +439,8 @@ class Component extends EventEmitter {
 		}
 		const instance = new Ctor(config, false);
 
-		if (window.__METAL_DEV_TOOLS_HOOK__) {
-			window.__METAL_DEV_TOOLS_HOOK__(instance);
+		if (window.__METAL_DEV_TOOLS_HOOK__ && window.__METAL_DEV_TOOLS_HOOK__.addRoot) {
+			window.__METAL_DEV_TOOLS_HOOK__.addRoot(instance);
```
jbalsas, line 443: Hey @bryceosterhaus, now that it's clear that we don't want to keep backwards compatibility, then this brings up my second concern, which is that every change in `metal-devtools` might need a corresponding change in devtools since this couples `metal.js` to a specific devtools implementation (now the `addRoot` method).

Are we doing this because you expect more hooks to be needed? I'm asking because I can see different ways to go about this.

**Only the render hook is needed**
Then we could keep this as is and simply refactor the devtools, so we keep `METAL_DEV_TOOLS_HOOK` as the entry point which takes care of calling any additional methods. This keeps `metal.js` agnostic of any `devtools` implementation.

**More than one hook is needed**
In this case, makes sense that a simple entry point won't be enough in the long run, so we could either do:

- Pass additional standard context information for the entry point to dispatch the actions
```javascript
window.__METAL_DEV_TOOLS_HOOK_(instance, {clazz: Component, method: 'render'})
```

- Stablish a proper convention after the method that is being hooked
```javascript
window.__METAL_DEV_TOOLS_HOOK_.render(instance);
// or
window.__METAL_DEV_TOOLS_HOOK_['Component::render'](instance);
```

Does any of that makes sense?

bryceosterhaus:
@jbalsas, yeah that makes sense. Ill keep it as is for now then and just modify devtools. That way metal shouldnt have to ever change and can always just be

```js
			if (window.__METAL_DEV_TOOLS_HOOK__) {
				window.__METAL_DEV_TOOLS_HOOK__(instance);
			}
```

Ill work on the changes to the devtools and keep you posted. Thanks for the input!

PR #200, opened 2017-01-18
Rename singleton as soyRenderer_ and add RENDERER_NAME property

jbalsas:


mairatma (approved):
You're right, this is necessary. Lgtm.

PR #203, opened 2017-01-27
Makes array.equal a little bit faster in case both arrays point to the same reference

brunobasto:


eduardolundgren:
Could you add a test for that?

brunobasto:
It doesn't really change any behavior, other than performance. Not sure how to write a test for that.

eduardolundgren:
It's for coverage purpose.

brunobasto:
@eduardolundgren, that makes total sense. I've added the test. Thx!

PR #204, opened 2017-01-31
Check for document fragments in toElement

julien:
Hi,

I did check the guidlines (and hope I didn't miss anything) but the `gulp lint` task did output a lot of warnings, before these changes, I wasn't sure if that was expected.

Thanks

### DIFF
--- a/packages/metal-dom/src/domNamed.js
+++ b/packages/metal-dom/src/domNamed.js
@@ -1,6 +1,6 @@
 'use strict';
 
-import { isDef, isDocument, isElement, isObject, isString, object } from 'metal';
+import { isDef, isDocument, isDocumentFragment, isElement, isObject, isString, object } from 'metal';
 import domData from './domData';
 import DomDelegatedEventHandle from './DomDelegatedEventHandle';
 import DomEventHandle from './DomEventHandle';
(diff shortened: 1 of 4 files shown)


eduardolundgren:
what happens if the document fragment is composed by multiple nodes and you invoke `toElement` on it?

yuchi:
You get the fragment itself. It works as an Element, a single one.

julien:
@eduardolundgren yes like @yuchi says, if you do something like this

```
var frag1 = document.createDocumentFragment();
var p = document.createElement('p');
var frag2 = document.createDocumentFragment();

var numElements = 1000;
for (var i = 0; i < numElements; i++) {
  var d = document.createElement('div');
  var s = 'child-' + i + 1;
  d.setAttribute('id', s);

  frag2.appendChild(d);
}
p.appendChild(frag2);
frag1.appendChild(p);

// assert.strictEqual(frag1, dom.toElement(frag1));
```

Then `dom.toElement(frag1)` returns you the `frag1` element, which is just another element.

yuchi:
…is just an Element-like API.

eduardolundgren:
I am familiar with the document fragment api, for a moment I though the pull request was converting a fragment to an element, therefore my question. It's clear now that it's just returning it. Thank you guys.

PR #207, opened 2017-02-07
Adds `internal` method to Config

mthadley:
I noticed `internal` was missing. Thanks!

### DIFF
--- a/packages/metal-state/src/Config.js
+++ b/packages/metal-state/src/Config.js
@@ -28,6 +28,17 @@ import validators from './validators';
  */
 const Config = {
 	/**
+	 * Adds the `internal` flag to the `State` configuration.
+	 * @param {boolean} required Flag to set "internal" to. True by default.
+	 * @return {!Object} `State` configuration object.
+	 */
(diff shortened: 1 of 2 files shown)


mthadley:
After thinking about this, I realized that this flag only makes sense for specific renderers. In this case , `IncrementalDOMRenderer`.  Even then, it's only useful for `Soy`. Compare that to the other methods like `required` and `value` which make sense for all renderers. We could still add this, but now I'm thinking it might make more sense to add some kind of `custom` method instead?

So it could work like this:
```js
MyComponent.STATE = {
  name: Config.string().value('test').custom('internal', true)
}
```

So custom would have a signature like `custom(name : string, value : any)`.

yuchi:
or…

```js
MyComponent.STATE = {
  name: {
    internal: true,
    ...Config.string().value('test')
  }
}
```

mthadley:
Yea, that's ok. Just thought I would add an alternative solution. If you like it, I'm fine with merging it. @eduardolundgren

PR #208, opened 2017-02-07
Exports Config from metal-soy

mthadley:
This exports `Config` from `metal-soy` in the same way that we do in `metal-jsx`.

Also, I wonder if going forward we should encourage using `Config` instead of doing something like `validator: core.isString`. `Config` seems to be more concise, and provide better feedback (in the form of warnings) when you provide invalid props. It's also more expressive and self-documenting when you need to validate enumerations (`oneOf`).


eduardolundgren:
For consistency I think it's a good addition. I will merge it, let me know if anyone has any concern.

/cc @mairatma

PR #217, opened 2017-03-06
Adds custom element registration for metal components

julien:
This is an initial implementation to add "WebComponents" support to metal.

Our initial idea was to be able to use metal components though custom elements, for example

```html
<my-component attr1="value1" attr2="value2"></my-component>
```

In order to do so, one would have to call the `defineWebComponent ` defined in metal's `core` page, like so

```javascript
import { core } from 'metal';
import Component from 'metal-component';

class MyComponent extends Component {
  // ...
}

core.defineWebComponent('my-component', MyComponent, ['attributes', 'to', 'be', 'observed']);
```

I would need some help from you to decide if this method should be defined in the core package, and also to set up travis (maybe these tests shouldn't run in "older" browsers?).

Any comment or feedback would be appreciated.

Thanks.

### DIFF
--- a/packages/metal-soy/package.json
+++ b/packages/metal-soy/package.json
@@ -22,6 +22,7 @@
     "metal"
   ],
   "dependencies": {
+    "@webcomponents/custom-elements": "^1.0.0-alpha.3",
     "html2incdom": "^0.1.1",
     "metal": "^2.6.4",
     "metal-component": "^2.7.0",
--- a/packages/metal-soy/test/Soy.js
+++ b/packages/metal-soy/test/Soy.js
@@ -1,5 +1,6 @@
 'use strict';
 
+import { core } from 'metal';
 import dom from 'metal-dom';
 import Component from 'metal-component';
 import { Events as EventsComponent } from './assets/Events.soy.js';
@@ -447,4 +448,70 @@ describe('Soy', function() {
 			assert.strictEqual(fn, Soy.toIncDom(fn));
 		});
 	});
+
+	describe('webcomponents', function() {
+
+		var el;
+
+		afterEach(function() {
+			if (el) {
+				document.body.removeChild(el);
+				el.component.dispose();
+			}
+		});
+
+		it('should not throw when creating a custom element', function() {
+      var tagName = createWebComponent();
+
+			var createEl = () => {
+				el = document.createElement(tagName);
+				return el;
+			};
+			var appendEl = () => document.body.appendChild(createEl());
+
+			assert.doesNotThrow(createEl);
+			assert.doesNotThrow(appendEl);
+		});
+
+		it('should dispatch events when attributes change', function() {
+
+			var tagName = createWebComponent();
+
+			el = document.createElement(tagName);
+			document.body.appendChild(el);
+
+			var fn = sinon.stub();
+
+			el.addEventListener('titleChanged', fn);
+
+			el.setAttribute('title', 'new title');
+			assert.strictEqual(1, fn.callCount);
+
+			el.setAttribute('non-existing', 'test');
+			assert.strictEqual(1, fn.callCount);
+		});
+
+	});
+
+	function createWebComponent() {
+		var tagName = `metal-test-component-${Date.now()}`;
+
+		class WebComponent extends Component { }
+		WebComponent.STATE = {
+			title: {value: 'default title'}
+		};
+
+		Soy.register(WebComponent, {
+			render: () => {
+				IncrementalDOM.elementVoid('div', null, [
+					'title', WebComponent.STATE.title.value
+				]);
+			}
+		});
+
+		core.defineWebComponent(tagName, WebComponent, ['title']);
+
+		return tagName;
+	}
+
 });
--- a/packages/metal/src/coreNamed.js
+++ b/packages/metal/src/coreNamed.js
@@ -289,3 +289,48 @@ export function isString(val) {
  */
 export function nullFunction() {
 }
+
+/**
+ * Register a custom element for a given metal component.
+ *
+ * @param {String} tagName The tag name to use for this custom element.
+ * @param {!function()} ctor Metal component constructor.
+ * @param {?Array=} observedAttrs The attributes to be "observed".
+ * @return {void} Nothing.
+ */
+export function defineWebComponent(tagName, ctor, observedAttrs) {
+  class CustomElement extends HTMLElement {
+    constructor() {
+      super();
+    }
+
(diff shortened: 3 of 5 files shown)


jbalsas:
Hey @julien, let's take a look together tomorrow! :)

jbalsas:
In any case, some tests are failing... could you double-check?

julien:
@jbalsas yes, of course

jbalsas:

yuchi:
Doubt: since it's a global registry shouldn't the caller register the components? Otherwise we are multiplying the risk of collisions.

jbalsas:
Hey @yuchi, do you mean something like:

```js
import { defineWebComponent } from 'metal';
import Component from 'metal-component';

class MyComponent extends Component {
  // ...
}

const CustomElement = createCustomElement(MyComponent/*, ['attributes', 'to', 'be', 'observed']*/);
window.customElements.define('my-component', CustomElement);
```

The way @julien set it up, the caller is calling a `define` helper, so I don't think it's a problem, but would love to hear your concerns :)

packages/metal-soy/test/Soy.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+import { core } from 'metal';
```
jbalsas, line 3: I think this should just be 

`import { defineWebComponent } from 'metal';`

jbalsas: But more importantly... why are the tests here instead of in the `metal` package?

packages/metal-soy/package.json
```diff
@@ -22,6 +22,7 @@
     "metal"
   ],
   "dependencies": {
+    "@webcomponents/custom-elements": "^1.0.0-alpha.3",
```
jbalsas, line 25: Should this be a `devDependency`?

packages/metal/src/coreNamed.js
```diff
@@ -289,3 +289,48 @@ export function isString(val) {
...
+
+    connectedCallback() {
+      let shadowRoot = this.attachShadow({mode: 'open'});
+      let opts = {};
+      if (Array.isArray(observedAttrs)) {
+        for (let i = 0, l = observedAttrs.length; i < l; i++) {
+          opts[observedAttrs[i]] = this.getAttribute(observedAttrs[i]);
+        }
+      }
+      this.component = new ctor(opts, shadowRoot);
```
jbalsas, line 319: This will likely give us a `A constructor name should not start with a lowercase letter` linting error...

packages/metal/src/coreNamed.js
```diff
@@ -289,3 +289,48 @@ export function isString(val) {
...
+
+/**
+ * Register a custom element for a given metal component.
+ *
+ * @param {String} tagName The tag name to use for this custom element.
+ * @param {!function()} ctor Metal component constructor.
+ * @param {?Array=} observedAttrs The attributes to be "observed".
+ * @return {void} Nothing.
+ */
+export function defineWebComponent(tagName, ctor, observedAttrs) {
```
jbalsas, line 301: Are we doing this for performance reasons or it is that isn't there a way to get the `observedAttrs` from the module definition?

jbalsas (changes requested):

packages/metal/src/coreNamed.js
julien: Ok thanks, I'll use uppercase.

yuchi:
From the example it wasn't clear if the registration happens in the context of creating the component or not.

The fact is, if I have two different version of metal-dropdown in my page I want to choose which one is registered.

packages/metal/src/coreNamed.js
julien: Ok, I will extract them from the component definition.

packages/metal-soy/test/Soy.js
julien: The tests are there because I wasn't sure if "DOM" related test should have been placed in the `metal` package, and `metal-soy` had everything needed. 

It might be a better idea to create a new `metal-custom-element` package and add all the related code in there.

julien:
@yuchi 

Sorry if I didn't understand correctly, but if you had two different versions of `metal-dropdown`  in your page wouldn't it be problematic anyway unless each version had a different `constructor` name? How would you do to distinguish one version from another?

julien:
There is also an issue I haven't disscused, which is the following:

The `CustomElement` class will create a new instance of the (metal) component that was passed to `defineWebComponent` method. 

The issue is that when this element is removed from the DOM, the component is not automatically "disposed". If you have any ideas on how we could detect this it would be great.

packages/metal/src/coreNamed.js
julien: Actually, when using 

```javascript
const observedAttrs = State.getStateStatic(Ctor);
```

To get the attributes we want to observe for the component definition, an error is thrown with the following message

```
Failed to execute 'define' on 'Custom Element Registry': Iterator getter is not callable.
```

I'm still trying to find a solution, but that's the reason why I added observedAttrs as an argument to the function.

julien: Actually using `metal-state` like this

```javascript
Object.keys(State.getStateStatic(Ctor));
```

Does the trick and avoids having to pass an extra argument, so that's a nice improvement.
Thanks

packages/metal-soy/package.json
julien: Yes it probably should.

Note that I was using this dependency, because when launching the tests, I was getting this error:

```
 Uncaught TypeError: Failed to construct 'HTMLElement': Please use the 'new' operator, this DOM object constructor cannot be called as a function. 
```
There is more info about this [here](https://github.com/babel/babel/issues/4480) and [here](https://github.com/babel/babel/issues/51169)

Also, I might try to use [this](https://github.com/github/babel-plugin-transform-custom-element-classes) babel plugin which hopefully solves the issue.

yuchi:
I believe the only way we can take is to dispose the component once `disconnectedCallback` is called. If the custom element is then reattached you need to create a new one.

This leads to state retainance issues between detach and attach cycles.

Should the new component retain the previous state? What happens when someone access the state after it has been disposed?

julien:
@yuchi Thanks. yes, I should have thought about `disconnectedCallback`.

julien:
Hi again,

I've done some changes, and added a new package called `metal-custom-element` which regroups this new funcionallity and tests, so I think it'd better to close this pull request and I will open a new one shortly.

PR #224, opened 2017-04-03
Adds custom element registration for metal components

julien:
This is the same feature as #217 but all the new code has been put in a new package called `metal-custom-element`. 

As usual, any feedback or improvements are welcome. 

Thanks

### DIFF
--- a/packages/metal-custom-element/src/customElement.js
+++ b/packages/metal-custom-element/src/customElement.js
@@ -29,7 +29,8 @@ export function defineWebComponent(tagName, Ctor) {
 				opts[observedAttrs[i]] = this.getAttribute(observedAttrs[i]);
 			}
 			this.component = new Ctor(opts, shadowRoot);
-			this.component.on('*', this.emit.bind(this));
+			this.componentEventHandler = this.emit.bind(this);
+			this.component.on('*', this.componentEventHandler);
 		}
 
 		attributeChangedCallback(attrName, oldVal, newVal) {
@@ -40,8 +41,8 @@ export function defineWebComponent(tagName, Ctor) {
 
 		disconnectedCallback() {
 			if (this.component) {
-				this.component.off('*', this.emit.bind(this));
-				this.component.detach();
+				this.component.off('*', this.componentEventHandler);
+				this.component.dispose();
 			}
 		}
 
--- a/packages/metal-custom-element/src/customElement.js
+++ b/packages/metal-custom-element/src/customElement.js
@@ -1,0 +1,57 @@
+import State from 'metal-state';
+
+/**
+ * Register a custom element for a given metal component.
+ *
+ * @param {String} tagName The tag name to use for this custom element.
+ * @param {!function()} Ctor Metal component constructor.
+ * @return {void} Nothing.
+ */
+export function defineWebComponent(tagName, Ctor) {
+
+	const observedAttrs = Object.keys(State.getStateStatic(Ctor));
+
+	class CustomElement extends HTMLElement {
+		constructor() {
+			super();
+		}
+
+		static get observedAttributes() {
+			return observedAttrs;
+		}
+
+		connectedCallback() {
+			let shadowRoot = this.attachShadow({
+				mode: 'open'
+			});
+			let opts = {};
+			for (let i = 0, l = observedAttrs.length; i < l; i++) {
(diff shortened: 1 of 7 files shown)


packages/metal-custom-element/src/customElement.js
```diff
@@ -0,0 +1,57 @@
...
+
+		attributeChangedCallback(attrName, oldVal, newVal) {
+			if (this.component) {
+				this.component[attrName] = newVal;
+			}
+		}
+
+		disconnectedCallback() {
+			if (this.component) {
+				this.component.off('*', this.emit.bind(this));
```
yuchi: This will **not** unbind the listener created at line 32 since `Function::bind` returns a new function everytime.

packages/metal-custom-element/src/customElement.js
```diff
@@ -0,0 +1,57 @@
...
+				this.component.detach();
```
yuchi: So we are still not calling `.dispose` on the component. What are the drawbacks of this?

julien: @yuchi I thought calling `detach` was correct, but if it needs to be `dispose` I have nothing against that, it's more because of my lack of knowlegde concerning metal than anything else.

packages/metal-custom-element/src/customElement.js
julien: good point

packages/metal-custom-element/src/customElement.js
yuchi: Calling detach would be perfect if we had a `garbageCollectedCallback` (which can't be added to the spec because it simply wouldn’t work).

Since we don‘t know when the custom element user will _actually_ throw away the element instance the only thing we can do IMHO is to destroy everything every time you get disconnected and recreate everything if connected again.

I’ll let @zenorocha @jbalsas @eduardolundgren chime in here.

julien:
Hi again, 

I'm trying to fix the tests that are failing and looking at the logs on travis, I see 

```shell
Chrome 48.0.2564 (Linux 0.0.0) ERROR
  Uncaught SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
  at /tmp/packages/metal-custom-element/node_modules/@webcomponents/custom-elements/src/custom-elements.js:24:0 <- /tmp/ee1ca899e5690f5cea35b86c25807372.browserify:5314
```

This file is out not ours and the version of Chrome is a bit outdated (from January 2016),
is there a way to use the latest "stable" version of Chrome?

Tests are also failing on:

  - IE 9.0.0 (Windows 7 0.0.0)

  - IE 11.0.0 (Windows 8.1 0.0.0) 

  - Edge 13.10586.0 (Windows 10 0.0.0)

  - Chrome Mobile 37.0.0 (Android 5.0.2) 

  - Safari 8.0.8 (Mac OS X 10.10.5)

And I was wondering if it was going to possible and reasonable to test "custom elements" in these browsers since it's quite a new (and still changing) feature, keeping in mind that we also have to solve [this](https://github.com/webcomponents/custom-elements#es5-vs-es2015) issue and that is why we're using the [native-shim](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js)

There are also tests failing on:

 - Firefox 45.0.0 (Linux 0.0.0) 

But from the looks of it, it seems that those errors concern `Soy` related funcionality so I wanted to know if we had already encountered this problem before and if we had been able to solve it?

Any help is welcome,
Thanks

julien:
Hi again,

I've been trying to fix the tests looking at the logs on travis but still no luck.

After having a look at the caniuse website for [custom element support](http://caniuse.com/#search=custom%20elements) I'm wondering if it's the right time to implement this feature since custom elements aren't available in most browsers except from Chrome.

Any thoughts or ideas are welcome.

julien:
I'm closing this one because it's getting too confusing to review and the feature is not yet completed:

- The custom element's [lifecycle callbacks](https://html.spec.whatwg.org/multipage/scripting.html#concept-custom-element-definition-lifecycle-callbacks) probably need some thinking.

- I'd like to try out [Web Component Tester](https://github.com/Polymer/web-component-tester) which seems to be an interesting tool for this feature.

Thanks again for your time.

jbalsas:
Thanks, Julien! 

We need to have this in place so people can opt in and start using it if it fits their needs ☺️

PR #225, opened 2017-04-04
Adds support for SoyToIncrDomSrcCompiler from https://github.com/goog…

jbalsas:
…le/closure-templates/commit/7dae5a1a2b6868da939e7643741b67db1610f190


jbalsas:
Hey @eduardolundgren, this is actually good to merge!

The problem is that in this project we're still using `metal-tools-soy@2.4.3`, so when I update, I hit this change: https://github.com/metal/metal-tools-soy/commit/9e5163d4040ff367af92f2b1a33deaf6c70cc428

Whenever we want to update the compiler version, we just need to also change https://github.com/metal/metal.js/blob/master/gulpfile.js#L100 and we're good to go!

Issue #209, opened 2017-02-08
hasClassWithNative_ throws an error when trying to check for several classes


antoniopol06:
The [`hasClass`](https://github.com/metal/metal.js/blob/master/packages/metal-dom/src/domNamed.js#L327)  method throws an error if it is invoked with a list of several classes when `classList` is available, for example:

```javascript

import {dom} from 'metal-dom';

let el = document.createElement('div');
el.classList.add('foo');
el.classList.add('bar');

dom.hasClass(el, 'foo bar'); // throws 
```

But, if `classList` is not available it works without errors.

fernandosouza:
Hello, @antoniopol06. Can you share the error message?

fernandosouza:
It's probably the token native error message alerting that the white space is not allowed. `Uncaught DOMException: Failed to execute 'contains' on 'DOMTokenList': The token provided ('foo bar') contains HTML space characters, which are not valid in tokens.`

We try to follow the DOM native API but our implementation of `hasClass` method is producing 2 different behaviours. We are going to look into it. Thank you.

eduardolundgren:
The hasClass only accepts one class per time, this matches the native behavior of classList contains.

fernandosouza:
Hey, @eduardolundgren. `hasClass` follow the native behaviour. But the method also has a fallback for browsers that don't have support for `classList` ([hasClassWithoutNative_](https://github.com/metal/metal.js/blob/master/packages/metal-dom/src/domNamed.js#L353)), and this method uses `indexOf()` to find the given class name inside `element.className`. In this case, you can pass multiple class names separated by space.

This is what @antoniopol06 has reported.

In case to follow the same behaviour, the fallback method should also throw an error when it receives more than one class name.

eduardolundgren:
Got it. Yes they should behave the same (as the native one).

fernandosouza:
Correcting what I said before, throws an error seems not be the right approach for this. Once [`tokenList.contains()`](https://dom.spec.whatwg.org/#dom-domtokenlist-contains) specification says: `The contains(token) method, when invoked, must return true if token is in tokens, and false otherwise.`

So, it says nothing about throws an error if white spaces are passed. And throws an error seems to be a Chrome implementation only.

In this case, we should return false if white spaces are passed.

jbalsas:
Fixed in https://github.com/metal/metal.js/pull/231!

PR #231, opened 2017-04-30
Issue/209

p2kmgcl:
These commits solve the report described con #209, however, there are three things pending before finishing:

- I only have passed the tests on latest Chrome and Firefox (for Linux). Checking the MDN documentation I have found that DOMException is a common type of exception implemented by Chrome, Firefox, Edge and IE (links below), buy I have no information about Safari implementation.
- I thought about another way of implementing hasClassWithNative_, and it's returning false if the given className has an space (same than hasClassWithoutNative_). Personally I prefer relying on the native implementation and then catching the error, but I am not sure what kind of behaviour you prefer.
- For checking the coverage, I temporary inverted the "has native implementation", because I have no browser without a native implementation. Maybe you have some kind of browserstack for the testing.

MDN DOMException docs: https://developer.mozilla.org/en-US/docs/Web/API/DOMException
MSDN DOMException docs: https://msdn.microsoft.com/es-es/library/ff974354(v=vs.85).aspx

### DIFF
--- a/packages/metal-dom/src/domNamed.js
+++ b/packages/metal-dom/src/domNamed.js
@@ -340,7 +340,8 @@ export function hasClass(element, className) {
  * @private
  */
 function hasClassWithNative_(element, className) {
-	return element.classList.contains(className);
+	return className.indexOf(' ') === -1 &&
+		element.classList.contains(className);
 }
 
 /**
@@ -351,7 +352,8 @@ function hasClassWithNative_(element, className) {
  * @private
  */
 function hasClassWithoutNative_(element, className) {
-	return (` ${element.className} `).indexOf(` ${className} `) >= 0;
+	return (` ${element.className} `).indexOf(` ${className} `) >= 0 &&
+		className.split(' ').length === 1;
 }
 
 /**
--- a/packages/metal-dom/test/dom.js
+++ b/packages/metal-dom/test/dom.js
@@ -150,6 +150,16 @@ describe('dom', function() {
 			dom.toggleClasses(element, 'adipiscing elit lorem sit consectetur');
 			assert.strictEqual('ipsum dolor amet adipiscing elit consectetur', element.className);
 		});
+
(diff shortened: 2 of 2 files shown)


packages/metal-dom/src/domNamed.js
```diff
@@ -340,7 +340,18 @@ export function hasClass(element, className) {
  * @private
  */
 function hasClassWithNative_(element, className) {
-	return element.classList.contains(className);
+	try {
```
eduardolundgren: try/catch won't be optimized in browsers, would be better to check for presence of `element.classList`.

p2kmgcl:
Ok, I have updated my PR and:

- These https://travis-ci.org/metal/metal.js/builds/227532190#L2322 "Error: Warning" were already on master, and do not make de tests fail, so I think I may ignore that.
- Can I have some kind of stack trace of https://travis-ci.org/metal/metal.js/builds/227532190#L2394? No idea of what code causes that.

jbalsas:
Hey @Robert-Frampton, this should be good to go. Care to take a look and merge if all seems fine? 

Thanks! ☺️

robframpton:
@p2kmgcl 

Why not remove the try catch block and check for whitespace in `_hasClassWithNative` like you are in `_hasClassWithoutNative`?

```javascript
function hasClassWithNative_(element, className) {
	return className.indexOf(' ') < 0 && element.classList.contains(className);
}
```

That way there is no error to catch.

p2kmgcl:
@Robert-Frampton you are totally right. I will modify the PR and notify you in a couple of days. Thanks for the correction :)

p2kmgcl:
@Robert-Frampton ready to go

robframpton:
LGTM.

brunobasto:
Hey guys, should we close https://github.com/metal/metal.js/issues/209 as well?

jbalsas:
Yeah, done! 😉

Issue #227, opened 2017-04-17
Config.instanceOf is not validating


AngeloYoun:
Setting config.instanceOf on a component state or prop doesn't properly validate. When adding logs to the function, the returned function is never ran.

```
Class Component extends JSXComponent{}

Component.PROPS = {
    foo: Config.instanceOf(class Foo{})
}

<Component foo={'hello'} />
```

does not throw an error.

AngeloYoun:
Thanks :)

PR #233, opened 2017-05-04
Fix for #227

bryceosterhaus:
We need to explicitly handle the use case for each validator since they all work differently. Validators that accept arguments all do something a little different. Previously, any validator that took an argument was never throwing an error.

### DIFF
--- a/packages/metal-state/src/Config.js
+++ b/packages/metal-state/src/Config.js
@@ -110,12 +110,82 @@ function mergeConfig(context, config) {
 	return obj;
 }
 
-// Add all validators to `Config`.
-const fnNames = Object.keys(validators);
-fnNames.forEach(
-	name => Config[name] = function() {
+/**
+ * Adds primitive type validators to the config object.
+ * @param {string} name The name of the validator.
+ */
+function setPrimitiveValidators(name) {
+	Config[name] = function() {
 		return this.validator(validators[name]);
-	}
-);
+	};
+}
+
+setPrimitiveValidators('any');
+setPrimitiveValidators('array');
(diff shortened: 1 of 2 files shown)


packages/metal-state/src/Config.js
```diff
@@ -110,12 +110,82 @@ function mergeConfig(context, config) {
...
+ */
+function setPrimitiveValidators(name) {
+	Config[name] = function() {
 		return this.validator(validators[name]);
-	}
-);
+	};
+}
+
+setPrimitiveValidators('any');
```
mthadley, line 123: Would it make more sense to make these (and `shapeOf`, `oneOfType`, etc.) just be methods on the `Config` as defined above? As oppose to declaring the object, and then mutating it down here over and over.

The implementation of each method could still just call some helper for the behavior that is shared.

Another idea would be to declare another object, something like `ConfigValidators`, and just merge the two objects at the end.

eduardolundgren: Regardless this proposal the end behavior should be the same, correct? If so this pull looks good to me and then if you want you can send a refactor for the internals.

mthadley: Yes, it would be the same.

It would just make this file much easier to read, since all methods on `Config` would be explicitly declared in the same way.

eduardolundgren: Agreed. Merging this one and refactoring is welcome. Thx.

bryceosterhaus: :thumbsup: I will go ahead and refactor and send another PR.

PR #234, opened 2017-05-04
Refactors for issue #227 and pr https://github.com/metal/metal.js/pull/233

bryceosterhaus:
@mthadley, what do you think about this?

### DIFF
--- a/packages/metal-state/src/Config.js
+++ b/packages/metal-state/src/Config.js
@@ -28,59 +28,77 @@ import validators from './validators';
  */
 const Config = {
 	/**
+	* An object that contains a validator function.
+	* @typedef {!Object} ConfigWithValidator
+	*/
+
+	/**
 	 * Function that creates `State` object with an `any` validator.
+	 * @return {ConfigWithValidator} `State` configuration object.
 	 */
 	any: setPrimitiveValidators('any'),
 
 	/**
 	 * Function that creates `State` object with an `array` validator.
+	 * @return {ConfigWithValidator} `State` configuration object.
 	 */
 	array: setPrimitiveValidators('array'),
 
 	/**
 	 * Function that creates `State` object with an `arrayOf` validator.
+	 * @param {ConfigWithValidator} stateConfig `State` configuration object
+	 * @return {ConfigWithValidator} `State` configuration object.
 	 */
 	arrayOf: setNestedValidators('arrayOf'),
 
 	/**
 	 * Function that creates `State` object with a `bool` validator.
+	 * @return {ConfigWithValidator} `State` configuration object.
 	 */
 	bool: setPrimitiveValidators('bool'),
 
 	/**
 	 * Function that creates `State` object with a `func` validator.
+	 * @return {ConfigWithValidator} `State` configuration object.
 	 */
 	func: setPrimitiveValidators('func'),
 
 	/**
 	 * Function that creates `State` object with an `instanceOf` validator.
+	 * @return {ConfigWithValidator} `State` configuration object.
 	 */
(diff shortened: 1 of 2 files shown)


mthadley:
🎉 Very nice! Just one more nitpick. We have docs for `oneOfType` and `shapeOf`, but is there any reason not to also document the rest of them? They are still methods on `Config`, after all.

packages/metal-state/src/Config.js
```diff
@@ -27,6 +27,41 @@ import validators from './validators';
...
+	 */
+	oneOfType(validatorArray) {
+		validatorArray = validatorArray.map(
+			configObj => configObj.config.validator
+		);
+
+		return this.validator(validators.oneOfType(validatorArray));
+	},
+
+	/**
```
mthadley, line 111: Missing a description here.

packages/metal-state/src/Config.js
```diff
@@ -27,6 +27,41 @@ import validators from './validators';
...
+	arrayOf: setNestedValidators('arrayOf'),
+	bool: setPrimitiveValidators('bool'),
+	func: setPrimitiveValidators('func'),
+	instanceOf: setExplicitValueValidators('instanceOf'),
+	number: setPrimitiveValidators('number'),
+	object: setPrimitiveValidators('object'),
+	objectOf: setNestedValidators('objectOf'),
+	oneOf: setExplicitValueValidators('oneOf'),
+
+	/**
```
mthadley: Missing a description here.

mthadley (changes requested):

bryceosterhaus:
Reason I didnt add docs for the rest of them is because the functions called are documented. I can add docs to everything, just wasnt sure.

bryceosterhaus:
Updated docs

mthadley:
@bryceosterhaus  I see. However, that doesn't mean that those methods are documented.

As an example, if you were to generate a documentation site from this file, those methods would not be documented.

bryceosterhaus:
Added separate commit to further document methods in config.

mthadley:
👍 👍 Really nice @bryceosterhaus 😍 😍 

Because of the usage for the more advanced validators is a little different, I'm hoping those extra docs will make it more obvious:

```js
MyComponent.STATE = {
  user: Config.shapeOf({
    /**
     * We actually have nested Config objects here, but only certain objects
     * make sense. In other words, you can't have `Config.value('Bryce')` here.
     * But `ConfigWithValidator` should be a nice hint.
     */
    name: Config.string(),
    age: Config.number()
  })
}
```

yuchi:
Could you update the title of the issue? I had a mini heart attack thinking it would be a Metal refactor :)

bryceosterhaus:
hahah yes I will update it, definitely don't want a full metal refactor

jbalsas:
Hey @bryceosterhaus, @mthadley, is this ready to be merged? I can do so if you both are happy with it.

bryceosterhaus:
@jbalsas lgtm

mthadley:
@jbalsas I think it is good to go.

packages/metal-state/src/Config.js
```diff
@@ -28,6 +28,104 @@ import validators from './validators';
  */
 const Config = {
 	/**
+	* A number, or a string containing a number.
```
jbalsas: I guess this description isn't really what a `ConfigWithValidator` is?

jbalsas (changes requested):

bryceosterhaus:
@jbalsas, updated the description of `ConfigWithValidator`

PR #237, opened 2017-05-12
Rename IncrementalDomAop.js to incremental-dom-aop.js

robframpton:


eduardolundgren:
LGTM

jbalsas:
Merged!

Hey @eduardolundgren, I copied this pattern from [`SoyAop`](https://github.com/metal/metal.js/blob/master/packages/metal-soy/src/SoyAop.js), so I guess we should rename that one as well...

jbalsas:
Actually, at least [`Config`](https://github.com/metal/metal.js/blob/master/packages/metal-state/src/Config.js) and [`iDOMHelpers`](https://github.com/metal/metal.js/blob/master/packages/metal-jsx/src/iDOMHelpers.js) seem to also have the wrong capitalization.

I guess we could just wait to rename them until we plan a major release, so unless you guys feel strongly about it, I'll leave these and `SoyAop` as they are and enforce this rule from now on :)

PR #241, opened 2017-05-26
Bug: elementClasses not being passed to child component

bryceosterhaus:
Hey, wanted to make sure that everyone agrees this is an issue before I look into fixing it.

I would expect that if a component returns another component at the top level, it should pass `elementClasses` down to that child component.

Reference to [higher-order-components](https://facebook.github.io/react/docs/higher-order-components.html)

### DIFF
--- a/packages/metal-incremental-dom/src/render/render.js
+++ b/packages/metal-incremental-dom/src/render/render.js
@@ -463,6 +463,18 @@ export function renderFunction(renderer, fnOrCtor, opt_dataOrElement, opt_parent
 function renderSubComponent_(tagOrCtor, config, opt_owner) {
 	const parent = getComponentBeingRendered();
 	const owner = opt_owner || parent;
+
+	const parentData = getData(parent);
+	const parentConfig = parentData.config;
+	if (!parentData.rootElementReached && parentConfig && isString(parentConfig.elementClasses)) {
+		let currentClasses = '';
+		if (isString(config.elementClasses)) {
+			currentClasses = `${config.elementClasses} `;
+		}
+
+		config.elementClasses = currentClasses + parentConfig.elementClasses;
+	}
+
 	const comp = getSubComponent_(tagOrCtor, config, owner);
 	updateContext_(comp, parent);
 
@@ -470,7 +482,6 @@ function renderSubComponent_(tagOrCtor, config, opt_owner) {
 	data.parent = parent;
 	data.owner = owner;
 
-	const parentData = getData(parent);
 	getChildComponents_(parentData).push(comp);
 	if (!config.key && !parentData.rootElementReached) {
 		config.key = parentData.config.key;
--- a/packages/metal-jsx/test/JSXComponent.js
+++ b/packages/metal-jsx/test/JSXComponent.js
@@ -306,6 +306,101 @@ describe('JSXComponent', function() {
 				done();
 			});
 		});
+
+		it('component.element and child.element should be the same', function() {
+			class ChildComponent extends JSXComponent {
+				render() {
+					return <div class="child"></div>;
+				}
+			}
+
+			class ParentComponent extends JSXComponent {
+				render() {
+					return <ChildComponent ref="child" />;
+				}
+			}
+
+			component = new ParentComponent();
+			var child = component.refs.child;
+			assert.strictEqual(component.element, child.element);
+		});
+
+		it('should receive elementClasses from parent component', function() {
+			class ChildComponent extends JSXComponent {
+				render() {
+					return <div class="child"></div>;
+				}
+			}
+
+			class ParentComponent extends JSXComponent {
+				render() {
+					return <ChildComponent />;
+				}
+			}
+
+			component = new ParentComponent({elementClasses: 'foo'});
+			assert.ok(dom.hasClass(component.element, 'child'));
+			assert.ok(dom.hasClass(component.element, 'foo'));
+		});
+
+		it('should not apply undefined class', function() {
+			class ChildComponent extends JSXComponent {
+				render() {
+					return <div class="child"></div>;
+				}
+			}
+
+			class ParentComponent extends JSXComponent {
+				render() {
+					return <ChildComponent />;
+				}
+			}
+
+			component = new ParentComponent({elementClasses: undefined});
+			assert.strictEqual(component.element.className, 'child');
+		});
+
+		it('should not create duplicate classes', function() {
+			class ChildComponent extends JSXComponent {
+				render() {
+					return <div class="child"></div>;
+				}
+			}
+
+			class ParentComponent extends JSXComponent {
+				render() {
+					return <ChildComponent />;
+				}
+			}
+
+			component = new ParentComponent({elementClasses: 'child'});
+			assert.strictEqual(component.element.className, 'child');
+		});
+
+		it('should pass elementClasses through higher order components', function() {
+			class GrandChildComponent extends JSXComponent {
+				render() {
+					return <div class="grandchild"></div>;
+				}
+			}
+
+			class ChildComponent extends JSXComponent {
+				render() {
+					return <GrandChildComponent elementClasses="child" />;
+				}
+			}
+
+			class ParentComponent extends JSXComponent {
+				render() {
+					return <ChildComponent elementClasses="parent" />;
+				}
+			}
+
+			component = new ParentComponent({elementClasses: 'foo'});
+			assert.ok(dom.hasClass(component.element, 'child'));
+			assert.ok(dom.hasClass(component.element, 'parent'));
+			assert.ok(dom.hasClass(component.element, 'grandchild'));
+		});
 	});
 
 	describe('shouldUpdate', function() {
--- a/packages/metal-soy/test/Soy.js
+++ b/packages/metal-soy/test/Soy.js
@@ -7,6 +7,7 @@ import { ComputedData as ComputedDataComponent } from './assets/ComputedData.soy
 import { ExternalTemplate as ExternalTemplateComponent } from './assets/ExternalTemplate.soy.js';
 import { Functions as FunctionsComponent } from './assets/Functions.soy.js';
 import { HelloWorld as HelloWorldComponent, templates as helloWorldTemplates } from './assets/HelloWorld.soy.js';
+import { HigherOrder as HigherOrderComponent } from './assets/HigherOrder.soy.js';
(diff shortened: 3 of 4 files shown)


mthadley:
In this case, I think both components would be sharing the same `element`. So that particular `element` should be receiving classes from multiple components. I think your test seems valid then 👍

jbalsas:
Hey @bryceosterhaus, I agree with @mthadley. Since it should be true that `component.element === child.element`, then it makes sense that the second one should have the `elementClasses` set properly in this case.

Are you going to give this a try?

bryceosterhaus:
@jbalsas yeah I have some time next week and I will take a look into it.

bryceosterhaus:
Just pushed some changes to this branch. I was able to find a way to fix the issue. However, I'm not sure this is the best place to implement it. Maybe someone with a better idea of the structure of metal.js might know.

It does pass the tests though :stuck_out_tongue_winking_eye:	  :tada:

eduardolundgren:
YUI Library had similar feature for Component inheritance. In Metal.js we've decided to not inherit **parent components** classes automatically for simplicity.

This pull request is a bit different though, it's not inheriting element classes from the **parent components**, instead it is inheriting from the **surrounding components**.

Based on [HOC](https://facebook.github.io/react/docs/higher-order-components.html) docs, seems like React doesn't inherit element classes automatically (correct me if I am wrong React folks).

```js
render() {
   // ... and renders the wrapped component with the fresh data!
   // Notice that we pass through any additional props
   return <WrappedComponent data={this.state.data} {...this.props} />;
}
```

Few considerations:

 * Is this change going to affect the behavior when inheriting it from a **parent component**? If not this should be fine to merge.
 * I am not sure if the patch is in the most appropriate place, @mairatma is the best person to double check.

Thanks @bryceosterhaus!

jbalsas:
I think in regards to YUI resemblance (@bryceosterhaus, @eduardolundgren, please correct me if I'm wrong), this might be something like having a Widget do:

```javascript
renderUI: function() {
    new AnotherComponent.render({
        boundingBox: this.get('boundingBox')
    });
}
```

Since we're handling `elementClasses` internally, I would expect to see it in the root element of whatever markup is generated from the component.

This behaviour is somewhat currently true in `soy` templates.

```soy
/**
 * 
 */
{template .render}
    {call .component /}
{/template}

/**
 * 
 */
{template .component}
    <div class="componentClass"></div>
{/template}
```

will render

```html
<div class="componentClass elementClasses"></div>
```

whereas

```soy
/**
 * 
 */
{template .render}
    <div>
        {call .component /}
    </div>
{/template}

/**
 * 
 */
{template .component}
    <div class="componentClass"></div>
{/template}
```

will render

```html
<div class="elementClasses">
    <div class="componentClass"></div>
</div>
```

This, of course, keeping in mind that for soy and server side rendering we need to explicitly add the `elementClasses` param if we want to avoid flickering :)

bryceosterhaus:
@eduardolundgren, you are correct in that React doesn't inherit classes automatically but that is because React doesn't have anything like `elementClasses`. If you want to pass a class into a component, you have to explicitly specify it as a prop. Since `elementClasses` is a special use case for us, I thought it would make sense to inherit since we are already doing this for [key](https://github.com/metal/metal.js/pull/241/files#diff-9c620b0e2dbc0335e959f81d6b3ed889R476). The reason I put it in patch, is because thats where we specified the logic of inheriting `key`. I am open to moving it somewhere else, just wasn't quite sure where that would be.

I think @jbalsas summed up the idea of the feature well in 

> Since we're handling elementClasses internally, I would expect to see it in the root element of whatever markup is generated from the component.

This way, the user always knows that `elementClasses` will do, and it is consistent from component to component.

yuchi:
My 2¢ citing @jbalsas on server side rendering: everything that breaks isomorphism should be avoided, unless explicit in the concept itself (for example an automatic param `isServerSide`).

eduardolundgren:
@yuchi Would you mind elaborating a bit more your last comment?

Regarding SSR, there are efforts happening for Metal isomorphism ability. We can continue this matter in a different thread.

yuchi:
@eduardolundgren Yeah sorry it was a little bit obscure.

From my very narrow point of view features that happen on the Metal.js side without being simple to port on the server-side rendering of Soy should be avoided.
In this specific case the feature on the server works if you manually pass `elementClasses` down the stream of templates. If you miss to do so you ① have flickering issues when components are rehydrated ② have different UIs when JS is enabled or not.

This is some behavior difficult to foresee from the (probably not very skilled) developer standpoint.

We either:

1. Keep things as they are and you are required to pass `elementClasses` to child templates, therefore:
    - docs should be **very** clear about this,
    - **every** Soy template in the Liferay Platform should follow this pattern (those are the main source of documentation and samples for Liferay developers)
    - we are imposing a tedious ceremony.
2. Make `elementClasses` work on templates rendered on Java too.

I repeat, this is my narrow PoV.

jbalsas:
Hey @yuchi, thanks a lot for your input! Great insight and to the point as usual 👍 

For now we are in the first scenario. I'll work next week on making sure this is very clear in the docs and double-check to make sure our soy templates are indeed following this pattern, which of course they should (as a matter of fact, metal-components do).

I think achieving the second point might be a bit complicated without making many assumptions and simplifications in the general Java Soy pipeline (this should not be an issue in the Node.js side of the story), but I think it is definitely worth at least exploring the options there, so we won't just abandon it. I'll let you know what we find in that sense.

mairatma:
Hey guys, sorry for the delay.

I also agree that this feature makes sense, `elementClasses` is already a special property we have, and the developer would surely expect it to apply to the component's element, regardless if that element is built by the component itself or a sub component.

The location for this fix feels correct as well, near where we do the same for `key`. The only thing I'd change is that this change is accessing `props` directly, which is a property that `metal-incremental-dom` shouldn't really know about, it's just a JSX specific feature. @bryceosterhaus doesn't it work by just setting the `config`? Maybe it would if you do this before calling `getSubComponent_`, instead of afterwards. I'm not sure though, my memory of this code is not as good as before anymore :)

bryceosterhaus:
@jbalsas @yuchi, I wasn't aware that in soy we had to handle this case differently. Evan helped clear things up for me by showing me how they are using it. After seeing the difference between how soy has to use it and jsx uses it, we were thinking it may not even make sense for `elementClasses` to be a feature.

> 1. Keep things as they are

This would probably be last choice in options. It is unintuitive knowing that `elementClasses` would be passed to a root element, but not a root component
 
> 2. Make elementClasses work on templates rendered on Java too.

I think this would be the ideal solution, that way we can have the same functionality across both soy and jsx without having the user understand any certain "gotcha" between the two.

I think there is also a 3rd option, and that would be to remove `elementClasses` altogether. This way we wouldn't have to manage these edge cases and users would have to manage their own classes. This is the way react handles classes and it wasn't too bad when we had to do that. Although I will way, the way that `elementClasses` merges classes on the root element was very helpful in jsx.


@mairatma, for some reason it didn't work by just changing config. I had also assumed that we shouldn't be messing with `props` in `metal-incremental-dom` but then I noticed that in that same file, props was already being used in multiple places, so I was a bit confused on that.

mairatma:
@bryceosterhaus oh, you're right, it must be confusing, but the `props` that are used in this file are actually for children config objects, which we pass down as the `children` property, but not as a component property in fact. The component property is unknown to this file, as not all components have it.

I think it may not have worked because it's being used after `getSubComponent`, and `elementClasses` needs to be already accessible by the component's constructor, which is not the case with `key`. So for this the logic should probably go before this `getSubComponent` call, and then it may work without accessing `props` directly.

mthadley:
@bryceosterhaus Just adding a quick thought. I think exploring option 2 is worthy endeavor either way. I'd like to see more work done on how we can take better advantage of the java side of the closure template system for our specific use cases.

`{2¢}`
In the case of Metal, Soy is not just being used as a basic templating system. It's the foundation of a component system. They have logic implicit in them. The templates are not just generating strings, but actually making calls into all kinds of arbitrary code (on the client side).

So I'd like to see more research in how we can take advantage of the server side API's that are offered. Things like their plugin system, or whatever else. Maybe we should also be more active on their repository, proposing more ideas.
`{/2¢}`

jbalsas:
Thanks everyone for chipping in!! 💰 

- We currently have this feature, which lots of projects depend on and this PR just makes it more consistent. **Let's land this** and move over the discussions to other threads. Please, @bryceosterhaus, could you take a look at @mairatma remarks to see if you can figure out a way to remove that dependency? Otherwise we can set some collaborative session next week during the gathering!

- We are almost done stabilising and improving our Soy integration in Liferay Portal, which is our main Java consumer for this. We will definitely start exploring more options, and of course, everyone is welcome to contribute 😉

bryceosterhaus:
@jbalsas, Just pushed updated logic to remove accessing props.

Thanks for the tip, @mairatma!

jbalsas:
Hey @bryceosterhaus, I'm not 100% sure what I was looking for, but I think I just spot a minor issue because we might not be checking if an `elementClasses` value has actually been passed down, so when that happens, we will be adding an `undefined` to the classes list.

```javascript
it('should not set undefined as an element class', function() {
	class GrandChildComponent extends JSXComponent {
		render() {
			return <div class="grandchild"></div>;
		}
	}

	class ChildComponent extends JSXComponent {
		render() {
			return (<div class="child">
				<GrandChildComponent />
			</div>);
		}
	}

	class ParentComponent extends JSXComponent {
		render() {
			return <ChildComponent />;
		}
	}

	component = new ParentComponent({elementClasses: 'foo'});
	assert.ok(dom.hasClass(component.element, 'child'));
	assert.ok(dom.hasClass(component.element, 'foo'));
	assert.ok(!dom.hasClass(component.element, 'grandchild'));
	assert.ok(!dom.hasClass(component.element, 'undefined'));
});
```

With this change, the test will go from this markup:

```html
<div class="child">
    <div class="grandchild" key="_metal_jsx_1"></div>
</div>
```

to this one:

```html
<div class="child undefined foo">
    <div class="grandchild" key="_metal_jsx_1"></div>
</div>
```

Could you please figure this one out before we push this through? 

Thanks!

bryceosterhaus:
@jbalsas, updated pr to include more test cases and fix the issue you noted

jbalsas:
Merged, with all tests `gulp test` and `gulp test:node` passing! 🎉 

Just moved stuff around. See [here](https://github.com/metal/metal.js/commit/49d7d2270f5a13deab7da6a1cea6abebeef45555)

Thanks!

Issue #205, opened 2017-02-02
Feature Proposal: otherProps


mthadley:
# Feature Proposal: `otherProps`

Hey everyone, so I had a feature proposal after seeing that we had a pattern that we both liked and saw ourselves repeating in probably all of our `metal-jsx` projects going forward. For now, I am calling it `otherProps` and it works something like this:

## The Problem

Let's say have a basic component like `Button` and it is defined like this:

```javascript.jsx
import Component, {Config} from 'metal-jsx';

class Button extends Component {
	render() {
		const {children, size} = this.props;

		return (
			<button class={`btn btn-size-${size}`}>
				{children}
			</button>
		);
	}
}

Button.PROPS = {
	size: Config.oneOf(['small', 'medium', 'large'])
};

export default Button;
```

It's pretty simple, but it allows use to create a component that behaves like a normal `button`, while exposing extra configuration that is specific to our app. However, we have a problem here when we want to start adding normal DOM attributes that we would expect to work with a `button` (or a `div`, `input`, etc.):

```javascript.jsx
class Button extends Component {
	render() {
		const {children, size} = this.props;

		return (
			<button 
				class={`btn btn-size-${size}`}
				disabled={this.props.disabled}
				onClick={this.props.onClick}
				onHover={this.props.onHover}
				onMouseOver={this.props.onMouseOver}
				autofocus={this.props.autofocus}
				form={this.props.form}
				formmethod={this.props.formmethod}
				formnovalidate={this.props.formnovalidate}
				name={this.props.name}
				type={this.props.type}
			>
				{children}
			</button>
		);
	}
}

Button.PROPS = {
	size: Config.oneOf(['small', 'medium', 'large']),
	disabled: Config.bool(),
	onClick: Config.func(),
	onHover: Config.func(),
	onMouseOver: Config.func(),
	autofocus: Config.bool(),
	form: Config.string(),
	formmethod: Config.string(),
	formnovalidate: Config.bool(),
	name: Config.string(),
	type: Config.string(),
	/* ...the possibilities go on... */
};
```

The API of your component will essentially keep growing endlessly as our use cases change and we require the ability to add basically any of the possible attributes in the HTML standard.

## The Solution

What we have instead done is use a helper function to simplify this. You can
see it copied [here](https://github.com/bryceosterhaus/metal-devtools/blob/beta-rewrite/frontend/lib/otherProps.js) to `metal-devtools`. So the previous example becomes this:

```javascript.jsx
[...]

Issue #127, opened 2016-06-10
Add a way to render html strings inside element in jsx components


mairatma:
It's currently possible to render html strings via Metal.js soy components, but not via jsx components. 

This was requested by @mthadley and @bryceosterhaus. Today they're going around this by using incremental dom directly, like [this](https://github.com/bryceosterhaus/liferay-plugins-ee/pull/554/files?diff=unified).

mthadley:
For ease of reference:

``` js
render() {
  IncrementalDOM.elementOpen('div', null, null, 'class', 'my-html-string-container');
  const node = IncrementalDOM.elementClose('div');
  node.innerHTML = '<div class="my-html-string">Hello World!</div>';
  return node;
}
```

mairatma:
This issue was moved to metal/metal#9

mairatma:
This issue was moved back.

bryceosterhaus:
Looks like the [easiest way](https://github.com/jridgewell/babel-plugin-transform-incremental-dom/issues/91#issuecomment-302567517) is to make that change in incremental-dom, but that would also affect soy. Should we do that, or find a way to make this change specifically on the JSX side of metal?

eduardolundgren:
@bryceosterhaus We might be able to to that for both JSX and Soy. If you have more details about how this can be achieved let me know.

mairatma:
Just FYI, at the time I remember I was thinking of just making a jsx component that could handle this, not of changing anything in metal.js itself. The component could even use [html2incdom](https://www.npmjs.com/package/html2incdom) to turn the inner html into incremental dom calls, or in the simplest case just do the same thing @mthadley posted here in its render function.

Not sure that's what you really need right now, just wanted to explain in case it might be of help :)

bryceosterhaus:
I was trying to avoid just creating a component for it because its just an additional thing to import. So I was hoping to implement it in [babel-plugin-transform-incremental-dom](https://github.com/jridgewell/babel-plugin-transform-incremental-dom/issues/91) since its just the  jsx side so i asked there and it was recommended to add it to our iDOM configuration but after testing that with our specific metal implementation there were issues with updates and patching. I tried testing an implementation it in iDOM by itself and seems to be that trying to set 'innerHTML' just doesn't work for our case.

So in the end, looks like the best route is to just create a component for this. I'll go ahead and create one and send a pr.

Issue #239, opened 2017-05-19
Repo Proposal: creating a metal-jsx-extras repository


bryceosterhaus:
I was looking into https://github.com/metal/metal.js/issues/205 and https://github.com/metal/metal.js/issues/127 and trying to figure out the best way to solve the two. After some research and talking with @mthadley, we thought it might be best to create a separate `metal-jsx-extras` repo that would contain extra(non-core) functionality for `metal-jsx`. This would be a place we could export a util function like `otherProps`(https://github.com/metal/metal.js/issues/205) and util components like `<DangerouslySetHTML />`(https://github.com/metal/metal.js/issues/127) since these are would be commonly used in along side of `metal-jsx`.

Usage:
```jsx
import Component from 'metal-jsx';
import {DangerouslySetHTML, otherProps} from 'metal-jsx';

class MyComponent extends Component {
  render() {
    const html = '<h1>Hello World!</h1>';

    return (
      <DangerouslySetHTML  {...otherProps(this)} html={html} />
    );
  }
}

export default MyComponent;
```

@eduardolundgren, @jbalsas  what are your thoughts on this? If it sounds like a good idea to you, I can go ahead and get it going.

jbalsas:
Hey @bryceosterhaus, after reading through #127 and #205, this comment by @mthadley particularly resonates:

> My first thought was that this could just be a separate module, completely outside of metal-jsx. However because it was such a strong pattern for us, I thought it might be good to propose it here first.

I guess the same might be said about the `DangerouslySetHTML` proposal, so **my preference would be to have those inside the current `metal-jsx` repo**. Inside an `extras` folder if you wish, but within the same package. We can always extract it later if needed.

What do you think?

bryceosterhaus:
I would be fine with putting it in the `metal-jsx` repo. However, the reason I would be okay with it is because I use both of those extras. For someone who may not use either, I could imagine that they wouldn't want those included with `metal-jsx`.

So if we think "everyone will use these extras in all of their projects" then I would say put it in the jsx repo. But if we anticipate that only 50% will use these extras for their project, then I would say that we should put it in a different place.

jbalsas:
I think you guys have leveraged `metal-jsx` more than anyone else, so it's just fair that you drive some of the guidelines. If that makes sense for you, it probably makes sense for more than just the 50% of the people.

Also, depending on the build process, you probably will be getting rid of that dead weight anyways...

mthadley:
[...]

PR #242, opened 2017-05-26
Add otherProps util and DangerouslySetHTML component

bryceosterhaus:
From discussion on https://github.com/metal/metal.js/issues/239

Fixes https://github.com/metal/metal.js/issues/127
Fixes https://github.com/metal/metal.js/issues/205

### DIFF
--- a/packages/metal-jsx/src/DangerouslySetHTML.js
+++ b/packages/metal-jsx/src/DangerouslySetHTML.js
@@ -1,0 +1,32 @@
+'use strict';
+
+import Component from './JSXComponent';
+import { Config } from 'metal-state';
+
+/**
+ * JSXComponent that renders html passed in.
+ * @class
+ */
+class DangerouslySetHTML extends Component {
+	/**
+	 * @return {Component}
+	 */
+	render() {
+		const {content, tag} = this.props;
+
+		IncrementalDOM.elementOpen(tag, null, null);
+
+		const node = IncrementalDOM.elementClose(tag);
+
+		node.innerHTML = content;
+
+		return node;
+	}
+}
+
+DangerouslySetHTML.PROPS = {
+	content: Config.string(),
+	tag: Config.string().value('span'),
+};
+
+export default DangerouslySetHTML;
--- a/packages/metal-jsx/src/JSXComponent.js
+++ b/packages/metal-jsx/src/JSXComponent.js
@@ -3,6 +3,7 @@
 import './iDOMHelpers';
 import { validators, Config } from 'metal-state';
 import Component from 'metal-component';
+import DangerouslySetHTML from './DangerouslySetHTML';
 import IncrementalDomRenderer from 'metal-incremental-dom';
 import JSXDataManager from './JSXDataManager';
 import JSXRenderer from './JSXRenderer';
@@ -38,4 +39,4 @@ JSXComponent.DATA_MANAGER = JSXDataManager;
 JSXComponent.RENDERER = JSXRenderer;
 
 export default JSXComponent;
-export { validators, Config, JSXComponent };
+export { DangerouslySetHTML, validators, Config, JSXComponent };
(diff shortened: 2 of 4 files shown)


bryceosterhaus:
Not sure why this is failing on travis...

jbalsas:
I've relaunched the tests, seemed like a network issue or something like that, all systems are green ☺️ 

I'll take a look at this on monday, but it looks good on first sight. In the meantime, maybe @mthadley  or @ethib137 who might be more familiar with the use case could chime in!

packages/metal-jsx/src/DangerouslySetHTML.js
```diff
@@ -0,0 +1,32 @@
...
+import Component from './JSXComponent';
+import { Config } from 'metal-state';
+
+/**
+ * JSXComponent that renders html passed in.
+ * @class
+ */
+class DangerouslySetHTML extends Component {
+  /**
+ * @return {Component}
```
mthadley: Indentation looks a little funny here.

packages/metal-jsx/src/otherProps.js
```diff
@@ -0,0 +1,24 @@
...
+ * Returns props that are not used or declared in the component.
+ * @param {Component} component Instance of metal-jsx component
+ * @return {Object} Object containing props
+ */
+export default function otherProps(component) {
+	const removeKeys = [...component.getDataManager().getPropsInstance(component).getStateKeys(), 'key', 'ref'];
+
+	const retObj = object.mixin({}, component.props);
+
+	removeKeys.forEach(
```
mthadley: I'm not sure if this really needs to change, but I can see most of the core library mostly prefers the C-style `for` loop over `forEach`. I'm guessing for performance? (The overhead of a function invocation for `n` items).

packages/metal-jsx/test/otherProps.js
```diff
@@ -0,0 +1,54 @@
...
+		};
+
+		assert.deepEqual(
+			otherProps(new TestComponent({baz: 'qux', foo: 'bar'})),
+			{baz: 'qux'}
+		);
+	});
+
+	it('should ignore key and ref', function() {
+		class TestComponent extends JSXComponent {
```
mthadley: I wonder if we can maybe start writing new tests completely in ES6? I mean, we are using `class` keywords already so...

mthadley (approved):
@jbalsas I added a few nitpicks, but all in all, this seems good. Thanks @bryceosterhaus!

mthadley:
@bryceosterhaus Oh, I just realized something. Maybe this could be a class method on `JSXComponent`?

```js
class MyComponent extends JSXComponent {
  render() {
    return <div ...this.otherProps() class="content" />;
  }
}
```

bryceosterhaus:
@mthadley, not a bad idea. Seems a little strange at first glance, but probably just because I am used to it being a separate function.

Also, for es6 in the unit tests, I decided to just follow the same patterns we had in our other tests

mthadley:
@bryceosterhaus I figured you were, which is fine. But maybe going forward we can start pushing for more idiomatic ES6. I'll defer to you and @jbalsas on this.

jbalsas:
Hey guys, thanks for working on this!

If `otherProps` is always meant to be used in the context of a `JSXComponent`, I also think it makes sense to make it part of the component directly, at least for now.

@bryceosterhaus, can you make that change and let me know? Regarding the tests, I'd love to start writing them in ES6 moving forward, but I wouldn't hold this PR for that, so no need to rewrite them right now.

Thanks!

bryceosterhaus:
@jbalsas @mthadley, just updated pr so that otherProps is on the component itself

jbalsas:
LGTM! @mthadley, one final check and I'll merge 😉

mthadley:
🎉  @jbalsas LGTM  🎉

eduardolundgren:
LGTM

eduardolundgren:
Could you add a test case for using `DangerouslySetHTML` from inside another component?

bryceosterhaus:
@eduardolundgren, added additional test case [here](https://github.com/metal/metal.js/pull/242/files#diff-d66c4d98a3d0bccc5855c5b99233541bR32)

ethib137:
LGTM

PR #256, opened 2017-09-11
Adds `writeOnce` method to state Config utils

robframpton:
There is a `writeOnce` property that can be set for state configuration that makes a property behave as read only, but our `Config` util doesn't have a shorthand method for it.


mthadley (approved):

jbalsas:
For some reason tests are failing... I'll try to get them to green before merging

jbalsas:
fixed travis and merged, thanks guys! 🎉

PR #257, opened 2017-09-13
Skip metal-custom-element tests when IE10 and below

robframpton:
From #230 

Hi everyone,

This pull request aims to bring Custom element support for Metal components.

A new "package" named `metal-custom-element` has been introduced for this feature.

The code has been tested on Chrome (`58.0.3029.81`), Firefox (`53.0`), but also IE11 (`11.0.0`) and Microsoft Edge (`14.14393.0`) with [ievms](https://github.com/xdissent/ievms) and the [karam-ievms-launcher](https://github.com/guillaumevincent/karma-ievms-launcher).

Here are a few notes that I wanted to share:

- Two new babel "plugins" have been introduced in order to "transpile" correctly 
  - https://github.com/github/babel-plugin-transform-custom-element-classes
  - https://www.npmjs.com/package/babel-plugin-transform-es2015-classes

- The [webcomponentjs](https://github.com/webcomponents/webcomponentsjs) polyfills are being used for browsers that don't natively support the different Web component APIs, and you'll notice that this dependency is being directly downloaded from the GitHub repository. The latest version is not published on [npm](https://www.npmjs.com/package/webcomponents.js) and we're not using [bower](https://bower.io/). An issue has been opened to get more information on when this package will be available through npm.

- At the moment, nothing has been done concerning the Custom element [adoptedCallback](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Custom_Elements#Custom_element_methods) method, mostly for simplicity and also because this callback is not supported by the polyfill (see: https://github.com/webcomponents/custom-elements#known-bugs-and-limitations) 

- Concerning the tests:

  - The polyfills used, are (or must be) doing something to `HTMLElement` (or `Node`) because they cause a test that was working to break

    This is the specific [test](https://github.com/metal/metal.js/blob/master/packages/metal-dom/test/dom.js#L191-L199), and here are some of the error messages in different browsers

    ```
    Firefox: Firefox 53.0.0 (Mac OS X 10.12.0) dom manipulation should append node list to parent element FAILED
            Argument 1 of Node.appendChild does not implement interface Node.
    ```

    ```
    Edge:   Edge 14.14393.0 (Windows 10 0.0.0) dom manipulation should append node list to parent element FAILED
            Error: No such interface supported
    ```

    ```
    IE11:   IE 11.0.0 (Windows 7 0.0.0) dom manipulation should append node list to parent element FAILED
            HierarchyRequestError
    ```

- Concerning **IE10**:

  - I have not been able to run the tests in IE10 despite multiple attempts. The "browserified" bundle loaded by Karma, launches an error in IE10, concerning `Object.setPrototypeOf`.
  
  - The polyfill used do no support **IE10** either.

  - I am trying to find a way to "skip" the tests, as well as not loading the bundled polyfills for IE10 and lower, but have not yet found a way. - Any help is welcome

As usual your ideas and feedback concerning improvements are welcome.

Thanks!


robframpton:
Hey @jbalsas @julien 

Here is an update to the web components branch. By default the components don't render using the shadow DOM, but it can be opted in using the `useshadowdom` attribute (all lowercase due to HTML spec).

```html
<my-metal-component useshadowdom="false" />
```

However, the tests are still failing due to the issue with the polyfill. I also noticed that the polyfill is 84kb minified. I'm hesitant about actually releasing this feature for two reasons.

1) That's a lot of extra JavaScript (84kb)

2) It's obviously affecting native DOM behavior in unintended ways (failing tests)

Maybe we can find another polyfill? Not sure what's out there.

jbalsas:
Hey, as a thought... why don't we just support this feature for evergreen browsers? Can we feature-check for `HTMLElement` and simply NOOP if not present?

That way we don't need to concern ourselves about the polyfill, and we can leave it up to users to see if it's worth their time?

robframpton:
So without any polyfill, the only browsers that should work are Chrome and Safari (which the tests are running against now). 

I'm fine with saying that's all we support, but I suppose we could provide instructions on how to include the polyfill in case someone wants to use it with other browsers as well, with the warning that it may lead to unintended consequences.

robframpton:
https://github.com/metal/metal.js/pull/258

Issue #259, opened 2017-09-21
Default values being overwritten by an undefined value. [metal-jsx]


bryceosterhaus:
When using default values for props, I would expect that if `undefined` is passed as a prop from the parent component, it would be ignored and the component would use its default prop value. However, this is not the case and currently it uses the undefined value.

Here is an example seen in metal, [Metal Example](https://jsfiddle.net/3eune52e/)

I think the react functionality for default props is what I would expect. [React Example](https://jsfiddle.net/hfqm34cp/1/).

kienD:
I think that this is a good idea.  I've run into this issue before.

bryceosterhaus:
@kienD do you use a certain workaround for this?

kienD:
@bryceosterhaus,

As of now, I'm just setting defaults value in the parent component so that the child component won't receive an undefined value.

robframpton:
Copying from #260 

I did some digging, and this behavior is an unintended side effect of f1e8847. That fix intended to ensure that validators are run for undefined values, in case isRequired was being used. However, it also makes it so that undefined will overwrite default values.

I ran @bryceosterhaus' tests against the commit right before f1e8847 and it passes.

Since this hasn't always been the behavior, I think we should go ahead and fix it. We can still run it against some larger code bases though, just to be safe.

jbalsas:
Merged into `develop` and scheduled for `2.14.0`!

bryceosterhaus:
@jbalsas should we be closing issues that are not released yet? Technically they are not "fixed" yet since it is not released. Only bring this up because I just spent 20 min debugging my code thinking this was fixed because it was closed :nerd_face:

jbalsas:
Hey @bryceosterhaus! Wasn't my message clear enough?

> Merged into develop and scheduled for 2.14.0!

We're trying to adopt a git-flow-like branching strategy, but that makes working with issues that are merged into the `develop` branch a bit complicated, because they stay open until merged to master... do you know how this gets handled in other projects? Any suggestion as to how to better communicate or organize this?

@Robert-Frampton, @brunobasto, thoughts?

bryceosterhaus:
I'm not super familiar with git-flow, so I can't really speak to that.

I'm not totally sure what other repos do or formal processes to follow for github issues. From my experience though, as a user of a project, I expect an issue to be closed/resolved once the issue is fixed and released for use. 

What I think of when I see issues:
**closed**: fixed and released
**open**: bug that exists in the latest release.

[...]

PR #260, opened 2017-09-21
Issue #259

bryceosterhaus:
Wanted to at least get a PR with a failing test. Also added a quick fix that seems to work and not break any tests, but I'm not positive this is the best way to go.

Also not entirely sure we want this functionality, this opens the table for discussion on it.

### DIFF
--- a/packages/metal-jsx/test/JSXDataManager.js
+++ b/packages/metal-jsx/test/JSXDataManager.js
@@ -374,5 +374,22 @@ describe('JSXDataManager', function() {
 			assert.strictEqual('defaultFoo', component.propsChanged.args[0][0].foo);
 			assert.strictEqual('foo', component.props.foo);
 		});
+
+		it('should use default prop value when `undefined` is passed', function() {
+			class TestComponent extends Component {
+			}
+
+			TestComponent.DATA_MANAGER = JSXDataManager;
+
+			TestComponent.PROPS = {
+				foo: {
+					value: 'defaultFoo'
+				}
+			};
+
+			component = new TestComponent({foo: undefined});
+
+			assert.strictEqual('defaultFoo', component.props.foo);
+		});
 	});
 });


jbalsas:
Hey @bryceosterhaus, thanks for sending this over!

Just wanted to point out that this is potentially a big breaking change if someone has code relying on `undefined` rather than the `default` value being used...

bryceosterhaus:
@jbalsas did we ever intend for a default value to be overridden by `undefined`? Seems strange that someone would pass it so that it would override. But yeah, it would break for those expecting an undefined value to pass through.

jbalsas:
It's highly unlikely that anyone is doing that. In any case, I'd say now this is a quirk much like many of the ones JS has, changing which might "_break the web_". 

In our case, I think @eduardolundgren might be the one with more exposure to this, so we could try running tests with this change over big metal codebases to check for regressions. Could we test Loop and Faro with this change locally to see if anything pops up?

bryceosterhaus:
Thats a good idea, I will see about trying it on Loop this week. As for the "fix" I did, it was just the first place I saw that it might work, more so wanted to get a proof of concept out there. It might require more effort for a proper fix.

jbalsas:
Hey @bryceosterhaus, any updates about this? 

We are planning to cut a release for next week (hopefully), so now should be the time! @Robert-Frampton, can you take a look at this?

bryceosterhaus:
@jbalsas The fix I added didn't work, so I removed that commit. Haven't had time to look into it any further.

robframpton:
I did some digging, and this behavior is an unintended side effect of f1e88470aec07e8b4c9f3e023c54c54837a3e4af. That fix intended to ensure that validators are run for `undefined` values, in case `isRequired` was being used. However, it also makes it so that `undefined` will overwrite default values.

I ran @bryceosterhaus' tests against the commit right before f1e88470aec07e8b4c9f3e023c54c54837a3e4af and it passes.

Since this hasn't always been the behavior, I think we should go ahead and fix it. We can still run it against some larger code bases though, just to be safe.

PR #264, opened 2017-10-02
importing metal-dom shouldn't immediately throw errors when running in Node.js environment

robframpton:
Hey @jbalsas 

This is just a step towards out of the box isomorphic rendering. With these changes you'll be able to have an import to `metal-dom` in a Node.js environment without it immediately throwing an error.

### DIFF
--- a/packages/metal-dom/src/features.js
+++ b/packages/metal-dom/src/features.js
@@ -32,6 +32,9 @@ class features {
 		const prefixes = ['Webkit', 'MS', 'O', ''];
 		const typeTitleCase = string.replaceInterval(type, 0, 1, type.substring(0, 1).toUpperCase());
 		const suffixes = [`${typeTitleCase}End`, `${typeTitleCase}End`, `${typeTitleCase}End`, `${type}end`];
+		if (!features.animationElement_) {
+			features.animationElement_ = document.createElement('div');
+		}
 		for (let i = 0; i < prefixes.length; i++) {
 			if (features.animationElement_.style[prefixes[i] + typeTitleCase] !== undefined) {
 				return prefixes[i].toLowerCase() + suffixes[i];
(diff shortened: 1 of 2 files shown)


packages/metal-dom/src/features.js
```diff
@@ -32,6 +32,9 @@ class features {
 		const prefixes = ['Webkit', 'MS', 'O', ''];
 		const typeTitleCase = string.replaceInterval(type, 0, 1, type.substring(0, 1).toUpperCase());
 		const suffixes = [`${typeTitleCase}End`, `${typeTitleCase}End`, `${typeTitleCase}End`, `${type}end`];
+		if (!features.animationElement_) {
+			features.animationElement_ = document.createElement('div');
```
jbalsas, line 36: Hey @Robert-Frampton, won't this still throw on usage on the server?

robframpton: Only if `checkAnimationEventName_` actually get's invoked, but it should never get invoked on the server. That's why the custom events are only being registered if it's [not server side](https://github.com/metal/metal.js/pull/264/files/f0bf96f991e032259022643c0e9dc2dd3d708a0e#diff-6bf9e779ca0474beb7faa9730d439844R50).

jbalsas: Yeah, I was thinking about other usages not protected here, like https://github.com/metal/metal-anim/blob/master/src/Anim.js#L60

robframpton: I think we might just need to state that some packages are not supported in server side rendering. All of the methods in metal-anim require DOM elements, so you really shouldn't be invoking on the server anyways.

robframpton: Also, even if a component implemented metal-anim in some way, it probably isn't going to be invoked by `renderToString`.

PR #265, opened 2017-10-02
Add simple shim so that importing metal-soy does not throw error when importing in Node.js environment

robframpton:
See https://github.com/metal/metal.js/pull/264

### DIFF
--- a/packages/metal-soy/src/Soy.js
+++ b/packages/metal-soy/src/Soy.js
@@ -1,5 +1,8 @@
 'use strict';
 
+// Must be imported before `html2incdom`;
+import './dom_shim';
+
 import 'metal-soy-bundle';
 import { ComponentRegistry } from 'metal-component';
 import { isFunction, isObject, isString, object } from 'metal';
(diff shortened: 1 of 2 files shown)


jbalsas:
Hey @Robert-Frampton, for this... we are pretty much working around https://github.com/mairatma/html2incdom, specially because of how it includes `HTMLParser`, iirc.

I'd like to keep `html2incdom` under control, moving it to the metal org. With that in mind... does it make sense to keep the existing one? Could we do it in a way that it uses modules properly and doesn't rely on globals? Mabye we could even use a different `HTMLParser` that we didn't need to maintain...

What do you think?

robframpton:
Hey @jbalsas 

I agree that we should be maintaining that functionality. What do you think about adding this API to `metal-incremental-dom` rather than creating a new package?

jbalsas:
That sounds lovely, @Robert-Frampton! 😍

robframpton:
I've sent https://github.com/metal/metal.js/pull/274 which copies `html2incdom` into `metal-incremental-dom`. It also _slightly_ modifies the thirdparty html parser so that it just exports the parser rather than attaching it to the `window`.

We might be able to replace it altogether with another parser, but I think we should go with this for now to avoid regressions.

jbalsas:
I assume #274 superseeds this one, right? Please reopen if we still need to merge this.

PR #266, opened 2017-10-02
Don't attempt to create element in Node.js environment

robframpton:
See https://github.com/metal/metal.js/pull/264

PR #269, opened 2017-10-03
Add tests for server side rendering of Soy and JSX components

robframpton:
**Note:** the tests on this branch will fail until the following branches are merged.

#264 
#265 
#266 

This branch adds the private `metal-isomorphic` package for the purpose of testing server side rendering of components. It's only intention is to add tests that can leverage from the lerna environment, which is why the `private` flag has been set to `true`.


jbalsas:
Sounds good, I added some comments on #264 and #265!

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+	"presets": ["es2015"]
```
jbalsas: I think `es2015` has been deprecated for a while in favor of `env`... could we try using it instead?

gulpfile.js
```diff
@@ -102,6 +102,12 @@ gulp.task('soy', function() {
 		.pipe(gulp.dest('packages/metal-soy/test'));
 });
 
+gulp.task('soy:isomorphic', function() {
```
jbalsas, line 111: Do we need gulp for this? Since we're invoking it from `npm run` can't we simply call `metalsoy`?

packages/metal-isomorphic/package.json
```diff
@@ -0,0 +1,17 @@
...
+	"repository": "https://github.com/metal/metal.js/tree/master/packages/metal-isomorphic",
+	"engines": {
+		"node": ">=0.12.0",
+		"npm": ">=3.0.0"
+	},
+	"dependencies": {
+		"metal-component": "^2.13.2",
+		"metal-jsx": "^2.13.2",
+		"metal-soy": "^2.13.2"
+	}
```
jbalsas, line 16: Aren't we missing `devDependencies` here? (I know it's probably picked by lerna, but it feels weird, not sure if that's how it's supposed to be :) )

packages/metal-isomorphic/test/isomorphic.js
```diff
@@ -0,0 +1,22 @@
+import Component from 'metal-component';
+import MyComponent from './fixtures/MyComponent';
+import MyJSXComponent from './fixtures/MyJSXComponent';
+import { assert } from 'chai';
```
jbalsas, line 4: Wouldn't we rather use `jest` snapshots here? Since we're already using it in other places and we won't be testing this on browsers... that way we have the same test dep everywhere.

jbalsas (changes requested):

gulpfile.js
robframpton: That's what I did at first, unfortunately this repo still uses an older version of `metal-tools-soy`, and updating it to the version that included the cli caused some issues.

Once this is merged I'd like to update it, resolve the issues, and remove whatever gulp tasks we can.

packages/metal-isomorphic/package.json
robframpton: Well, since the tests aren't actually run from this package, it will only add bloat to `lerna bootstrap` since the `devDependencies` in this package wouldn't be used. I can see why it looks a little confusing though, do you want to add `mocha` and the other test dependencies here?

packages/metal-isomorphic/test/isomorphic.js
robframpton: Sure we can do that.

robframpton: Do you think we can make that improvement down the line? There are a lot of places in this repo that we could be using jest snapshots, it would be nice to do that migration all at once under a separate issue.

jbalsas: Sure, we can do that

packages/metal-isomorphic/package.json
jbalsas: Sounds good, if that's what we're also doing in other projects... we need to converge all the monorepos' infrastructure :)

gulpfile.js
jbalsas: fair enough!

jbalsas:
This LGTM, then, just waiting on https://github.com/metal/metal.js/pull/274 to retest.

jbalsas (approved):

robframpton:
Waiting on #274 to rebase this branch, as it will cause more conflicts.

robframpton:
Note: I added 4c2a707 since we decided to not go with #265. Apparently that shim was also allowing `Component` to work server side.

jbalsas:
Awesome!!! All is green!!! 🎉 

As I mentioned, my only concern right now is that testing jumps from 6-8 minutes to 13-15 on this PR, so I'd like to see if we can figure out why. Could you take a look?

robframpton:
It's retesting right now, but I suspect it's because Edge is hanging for some reason.

robframpton:
@jbalsas Edge didn't hang in the retest, and the time was significantly shorter.

jbalsas:
Merging then!! 🎉

Issue #255, opened 2017-09-07
Improving Metal.js Lifecycle and Compatiblilty


mthadley:
Now that Metal.js is more mature, we can start thinking about both improving it's lifecycle features and also making it more compatible with knowledge from other existing frameworks. In particular, we want to look at moving over to lifecycle names that are also found in React. This would be similar to the relationship between [preact](https://preactjs.com/) and React.

So as an example, we could either alias or rename `attached` to `componentDidMount`. Or we could add new lifecycles, like `componentWillRecieveProps`.

@jbalsas @bryceosterhaus @eduardolundgren Moving this conversation here.

jbalsas:
Hey @mthadley, I've tried to gather some information here to see what we have and what we might need. Could you double check and expand on this?

|  | React.js | Metal.js |  |
| --- | --- | --- | --- |
| **Mounting** | [constructor](https://facebook.github.io/react/docs/react-component.html#constructor) | `constructor` | ✅ |
| | [render](https://facebook.github.io/react/docs/react-component.html#render) | `render` | ✅ |
| | [componentWillMount](https://facebook.github.io/react/docs/react-component.html#componentwillmount) | | ⚠️ |
| | [componentDidMount](https://facebook.github.io/react/docs/react-component.html#componentdidmount) | `attached` | ✅ |
| **Updating**| [componentWillReceiveProps](https://facebook.github.io/react/docs/react-component.html#componentwillreceiveprops) | | ⚠️ |
| | [shouldComponentUpdate](https://facebook.github.io/react/docs/react-component.html#shouldcomponentupdate) | `shouldUpdate` | ✅ |
| | [componentWillUpdate](https://facebook.github.io/react/docs/react-component.html#componentwillupdate) | | ⚠️ |
| | [componentDidUpdate](https://facebook.github.io/react/docs/react-component.html#componentdidupdate) | `rendered` | ✅ |
| **Unmounting** | [componentWillUnmount](https://facebook.github.io/react/docs/react-component.html#componentwillunmount) | | ⚠️ |
| **APIs** | [setState](https://facebook.github.io/react/docs/react-component.html#setstate) | `setState` | ✅ |
| | [forceUpdate](https://facebook.github.io/react/docs/react-component.html#forceupdate) | | ⚠️ |
| **Class Properties** | [defaultProps](https://facebook.github.io/react/docs/react-component.html#defaultprops) | | ⚠️ |
[...]

PR #270, opened 2017-10-06
Add willReceiveState and willReceiveProps lifecycle methods to Component and JSXComponent respectively

robframpton:
See https://github.com/metal/metal.js/issues/255


packages/metal-component/src/Component.js
```diff
@@ -172,6 +182,8 @@ class Component extends EventEmitter {
 	 */
 	attach(opt_parentElement, opt_siblingElement) {
 		if (!this.inDocument) {
+			this.emit('willAttach');
+			this.willAttach();
```
jbalsas, line 186: Hey @Robert-Frampton, based on [componentWillMount()](https://reactjs.org/docs/react-component.html#componentwillmount):

> `componentWillMount()` is invoked immediately before mounting occurs. It is called before render(), therefore setting state synchronously in this method will not trigger a re-rendering. Avoid introducing any side-effects or subscriptions in this method.

Can we assume the same with this?

packages/metal-component/src/Component.js
```diff
@@ -501,6 +527,9 @@ class Component extends EventEmitter {
 	 *     be called manually later to actually attach it to the dom.
 	 */
 	renderComponent(opt_parentElement) {
+		const firstRender = !this.hasRendererRendered_;
+		this.emit('willRender', firstRender);
+		this.willRender(firstRender);
```
jbalsas, line 532: Based on [componentWillUpdate()](https://reactjs.org/docs/react-component.html#componentwillupdate), the signature for this method should be:

```javascript
componentWillUpdate(nextProps, nextState)
```

Furthermore:

> `componentWillUpdate()` is invoked immediately before rendering when new props or state are being received. Use this as an opportunity to perform preparation before an update occurs. **This method is not called for the initial render**.

Finally, do we have the last note in mind? 

> `componentWillUpdate()` will not be invoked if shouldComponentUpdate() returns false.

jbalsas (commented):
Hey @Robert-Frampton, I've added a couple of notes about `willAttach` and `willRender`. Can you take a look at those?

Also, I kinda added a merge commit here by mistake. Could you send a new PR once you're ready? 👼 

Thanks!

packages/metal-component/src/Component.js
robframpton: Hey @jbalsas 

So an area where Metal is already different than React is that our `render` actually happens before `attach`. This is how the lifecycle is ordered on first render.

`created > willRender > rendered > willAttach > attached > willDetach > detached > disposed`

I think the only way we can mirror React completely is to do a breaking change and change the order so that attachment happens before render, but obviously we can't do that for `2.14`. Although we could move `willAttach` only, but it feels weird to me.

`created > willAttach > willRender > rendered > attached > willDetach > detached > disposed`.

robframpton:
https://github.com/metal/metal.js/pull/285

packages/metal-component/src/Component.js
robframpton: I update this in the new pull. I changed the name to `willUpdate` since it's closely related to `shouldUpdate` and receives the same arguments.

packages/metal-component/src/Component.js
jbalsas: That's fine for now... I think we are entitled to some little quirks of our own ;)

I think we just need to improve our [Component Lifecycle](https://metaljs.com/docs/guides/component-lifecycle.html) documentation so everything is clear :)

PR #272, opened 2017-10-06
Improve error message when using `oneOf` validation. Fixes #254

### LINKED ISSUE #254: Improve error message when using `oneOf` validation
When using `oneOf` validation for a Component property, I would like the error message to display the valid types.

This is what is currently displayed.

```
Error: Warning: Invalid state passed to 'type'. Expected one of given values. Passed to 'MyComponent'.
```

This is what I'd like to see.

```
Error: Warning: Invalid state passed to 'type'. Expected one of the following values: 'value1', 'value2', 'value3'. Passed to 'MyComponent'.
```

robframpton:


packages/metal-state/src/validators.js
```diff
@@ -185,6 +186,15 @@ function composeError(error, name, context) {
 }
 
 /**
+ * Composes an error message for Conifg.oneOf validator.
```
jbalsas: Possible typo: `Conifg.oneOf` -> `Config.oneOf`

jbalsas (changes requested):
Hey @Robert-Frampton, should we add a test for this?

robframpton:
We don't currently test the contents of any error messages, we only test that an error is returned or not given the input. We can add tests for this, but if we do we should probably add tests for the other use cases as well.

jbalsas:
Sounds good, just the typo, then 👍

diegonvs:
Hey @Robert-Frampton i was reviewing that tests and i've started a discussion here:
https://github.com/metal/metal.js/issues/277

PR #275, opened 2017-10-11
Check to see if bundle has been loaded already. Fixes #273

### LINKED ISSUE #273: metal-soy-bundle shouldn't be so fragile
In working on creating a lerna repo with multiple soy components in `packages`, I ran into the familiar error `Uncaught Error: Namespace "goog.string" already declared.`.

In most scenarios we're already deduping our bundles which is fine, but it shouldn't be a requirement for simply importing more than one Soy component into a bundle, and it's not actually easy to perfectly dedupe in some testing environments. We should continue to dedupe, and encourage others to do the same, but `metal-soy-bundle` shouldn't be so fragile.

If it's a file that can truly only be loaded once in the client, then why not set some kind of flag to indicate that the bundle has been loaded already? Then check that flag before trying to invoke it a second time?

[Here](https://github.com/metal/metal.js/blob/master/packages/metal-soy-bundle/gulpfile.js#L26) we are already prepending some logic to the file, if we add a simple snippet it should prevent those errors from being thrown.

```javascript
if (this.__METAL_SOY_BUNDLE_LOADED__) {
  return;
}
this.__METAL_SOY_BUNDLE_LOADED__ = true;
```

We could even show a warning that the file has been imported multiple times, so that we're still encouraging deduping.

I suppose we could also check to see if `goog.string` is defined, I guess I just like the idea of being more explicit so the intention of the check is more obvious.

@jbalsas @brunobasto 

Edit: [just so it's more clear](https://github.com/Robert-Frampton/metal.js/commit/eeb36e083db5e1a499c270560d108b3b6a5f6ccc)

robframpton:


jbalsas:
Hey @Robert-Frampton, this looks good in general, I just have one question. Even though I tweaked the `metal-soy-bundle` build, I quite never remember how it is setup and what is what 😂.

What is the difference between `metal-soy-bundle/build/bundle.js` and `metal-soy-bundle/lib/bundle.js`? Shouldn't we try to load `lib` twice instead of `lib/bundle.js` and `build/bundle.js`?

While we're at this... could we add some information in [metal-soy-bundle/README.md](https://github.com/metal/metal.js/blob/master/packages/metal-soy-bundle/README.md) clarifying for future reference?

robframpton:
Hey @jbalsas 

`build/bundle.js` is just the concatenated files, nothing more. That file is then transpiled which generates `lib/bundle.js`. 

The reason I loaded the `build/bundle.js` file in the test, is that Karma has some kind of mechanism for making sure a file is loaded only once, so I couldn't get it to load a second time when pointing to `lib/bundle.js`.

And yes, we can add some readme info on the build process :)

jbalsas:
I see... and couldn’t we trick Karma with some kind of cache querystring? &t=123?

robframpton:
That's what I tried actually, didn't have any affect on it.

robframpton:
Added a short description of the build process.

jbalsas:
Hey @Robert-Frampton, looks like this actually broke some tests when merged 😢. I'm trying to see why the merge tests are not running in Travis (only the PR ones).

In the meantime, I can revert this so we can process the others, or you can send a followup PR fixing it. What would you prefer? 🤷‍♂️

yuchi:
**Question:** does the solution to the issue at hand make impossible to have multiple Metal.js versions in the same page? _**Hint:** it’s going to happen when deploying modern bundles to portal built with different metal version._

jbalsas:
Hey @yuchi, that's actually a very interesting question indeed... I don't think it has ever been possible, because both `incr-dom` and `soy-bundle` leak to the global scope.

We're going to revert it in any case. We can rather approach this with that in mind. It would involve touching both `incr-dom` and `soy-bundle`, so we'll have to plan that for some later time...

PR #278, opened 2017-10-16
Updating tests on validators using assert.isTrue() to verify correct cases

diegonvs:


brunobasto:
Looks good. Luckily there were no underlying issue uncovered by the tests.

robframpton (approved):

PR #285, opened 2017-10-23
Rename willRender to willUpdate since it's related to shouldUpdate and move implementation to renderer

robframpton:


packages/metal-component/src/Component.js
```diff
@@ -650,6 +676,16 @@ class Component extends EventEmitter {
...
+
+	/**
+	 * Lifecycle. Fires before the component has been attached to the DOM.
+	 */
+	willAttach() {}
+
+	/**
+	 * Lifecycle. Fires before component is detached from the DOM.
+	 */
+	willDetach() {}
```
jbalsas, line 734: Should we also have an empty placeholder for `willUpdate`? I see we are actually checking down the line to see if `component.willUpdate` exists, but it feels a bit inconsistent with how we are treating every other lifecycle method.

The most similar case seems to be `rendered`, which is actually always used from the `component` instance `informRendered` wrapper method...

packages/metal-component/test/Component.js
```diff
@@ -94,6 +94,29 @@ describe('Component', function() {
 			assert.ok(comp.inDocument);
 		});
 
+		it('should run "willAttach" lifecycle method when the component about to attach', function() {
```
jbalsas: Missing verb? `when the component **is** about to attach`?

robframpton:
Just noticed this has merge conflicts. I'll rebase.

packages/metal-component/test/Component.js
```diff
@@ -94,6 +94,29 @@ describe('Component', function() {
...
+			class TestComponent extends Component {
+			}
+			sinon.spy(TestComponent.prototype, 'willAttach');
+			comp = new TestComponent();
+
+			assert.ok(comp.willAttach.calledBefore(Component.prototype.attached));
+			assert.strictEqual(1, comp.willAttach.callCount);
+		});
+
+		it('should emit "willAttach" lifecycle method when the component about to attach', function() {
```
jbalsas: Are we saying `emit ... method` in the rest of tests? maybe better to say `emit ... lifecycle **event** when the component **is** about to...`?

packages/metal-component/test/Component.js
```diff
@@ -142,6 +165,37 @@ describe('Component', function() {
 			assert.strictEqual(comp, comp.attach());
 		});
 
+		it('should run "willDetach" lifecycle method when the component about to detach', function() {
```
jbalsas: Missing verb? when the component **is** about to detach?

packages/metal-component/test/Component.js
```diff
@@ -142,6 +165,37 @@ describe('Component', function() {
...
+
+			assert.strictEqual(0, comp.willDetach.callCount);
+
+			comp.detach();
+
+			assert.ok(comp.willDetach.calledBefore(Component.prototype.detached));
+			assert.strictEqual(1, comp.willDetach.callCount);
+		});
+
+		it('should emit "willDetach" lifecycle method when the component about to detach', function() {
```
jbalsas: Are we saying emit ... method in the rest of tests? maybe better to say emit ... lifecycle **event** when the component **is** about to...?

jbalsas:
hey @Robert-Frampton, I'm finishing the review, hold on if you wish so you can do all at once!

packages/metal-incremental-dom/src/IncrementalDomRenderer.js
```diff
@@ -218,10 +222,25 @@ class IncrementalDomRenderer extends ComponentRenderer.constructor {
...
 		}
 	}
+
+	/**
+	 * Invokes component's "willUpdate" lifecycle method if applicable.
+	 * @param {!Component} component
+	 * @param {Object} changes
+	 */
+	willUpdate_(component, changes) {
+		if (!component.willUpdate || !component.wasRendered || !changes) {
```
jbalsas: See comment above... if we want to follow the `render` approach, we could provide a wrapping API for it like `component.informWillUpdate`. I'm fine keeping it here too :)

jbalsas: Also, from the calling code, it might not be needed to check for `!changes`, since it is always invoked within something that used `shouldUpdate(..., changes)` which already does that check?

packages/metal-jsx/src/JSXComponent.js
```diff
@@ -34,6 +34,18 @@ class JSXComponent extends Component {
...
 
 	/**
+	 * Fires before state batch changes. Provides hook point for modifying
+	 *     state.
+	 * @param {Event} event
+	 * @protected
+	 */
+	handleStateWillChange_(event) {
+		if (event.type !== 'state' && this.willReceiveProps) {
+			this.willReceiveProps(event.changes);
```
jbalsas: This is only for JSX, right? We still should document it somehow in the docs, as we do for the other lifecycle methods. I know we are likely updating the site docs with this, but it is fairly easy to miss on code inspection.

Also, same remarks as `willUpdate` about providing a default empty placeholder.

packages/metal-state/src/State.js
```diff
@@ -304,6 +304,8 @@ class State extends EventEmitter {
 	 */
 	emitBatchEvent_() {
 		if (!this.isDisposed()) {
+			this.context_.emit('stateWillChange', this.scheduledBatchData_);
```
jbalsas, line 310: Not related to this change in particular, but I'm wondering how can we document what events are emitted so one knows what to listen :)

jbalsas (changes requested):
Hey @Robert-Frampton, looks good to me in general, just some minor remarks and the rebase and we should be ready to go! 👍

packages/metal-incremental-dom/src/IncrementalDomRenderer.js
robframpton: Hey @jbalsas 

In that case we might want to do the same for `shouldUpdate`, right now it just checks for a `component.shouldUpdate` method from the [renderer](https://github.com/metal/metal.js/blob/master/packages/metal-incremental-dom/src/IncrementalDomRenderer.js#L191-L193). What do you think?

jbalsas: Yup, we probably should! 👍

robframpton:
Hey @jbalsas 

I made most of the changes you mentioned. I decided against adding an `informShouldUpdate` method and placeholder `shouldUpdate` method.

This is because it would be the only lifecycle method that needs to return a value, it just felt weird doing this...

```javascript
informShouldUpdate(...args) {
  return this.shouldUpdate(...args);
}

shouldUpdate() {
  return true;
}
```
None of the other `inform...` methods need to return anything back to the renderer, at that point in the contract the renderer is just informing the component that something happened and doesn't care what the component does.

We can still do it if you think this discrepancy is fine.

Edit: I should also note that if we add a placeholder that we'll have to change [this logic](https://github.com/metal/metal.js/blob/master/packages/metal-soy/src/Soy.js#L181) in the Soy component.

packages/metal-component/src/Component.js
```diff
@@ -27,13 +27,32 @@ import { EventEmitter, EventHandler } from 'metal-events';
  * class CustomComponent extends Component {
  *   created() {
  *   }
- *
+
```
jbalsas: Unintended deletion?

packages/metal-component/src/Component.js
```diff
@@ -27,13 +27,32 @@ import { EventEmitter, EventHandler } from 'metal-events';
...
+ *   willAttach() {
+ *   }
+ *
  *   attached() {
  *   }
  *
+ *   willReceiveState() {
+ *   }
+ *
+ *   // JSX components only
```
jbalsas: This comment here is a bit confusing... for a while I assumed everything below this point was related to JSX components only and not just this method... maybe place the comment inside? or make it more explicit, like `willReceiveProps is only available in JSX components`?

jbalsas (changes requested):
Hey @Robert-Frampton, looks good, just 2 final issues, (sorry for the nit-picking) and we're good to go!

No need to get crazy about the `forceUpdate` placeholder, let's see how this works and we can iterate later.

robframpton:
Sorry, what exactly do you mean in regards to a `forceUpdate` placeholder?

Edit: oh, I'm guessing you meant `shouldUpdate` ;)

jbalsas:
Sorry, I meant `shouldUpdate`. It was related to your [previous comment](https://github.com/metal/metal.js/pull/285#issuecomment-341468931). Sounds good to me 👍

robframpton:
Updated.

jbalsas (approved):

PR #288, opened 2017-10-31
Ensure that STATE configuration from static hints are properly configured with multiple levels of class inheritance

robframpton:


packages/metal-state/test/State.js
```diff
@@ -1041,6 +1041,32 @@ describe('State', function() {
...
+
+			var test = new Test({
+				key1: 'foo1'
+			});
+			assert.strictEqual('foo1', test.key1);
+
+			var child = new Child({
+				key2: 'foo2'
+			});
+			assert.strictEqual('foo1', test.key1);
```
jbalsas, line 1067: Shouldn't this be asserting `child.key1`?

jbalsas: Actually, if you change it to `child.key1`, the test will fail...

jbalsas (changes requested):
Hey @Robert-Frampton, I'm not sure the test is correct, and thus the solution might be incomplete.

Also, while we're at it, the `configStateFromStaticHint_` jsdoc says:

```javascript
/**
 * Adds state keys from super classes static hint `MyClass.STATE = {};`.
 * @param {Object.<string, !Object>=} opt_config An object that maps all the
 *     configurations for state keys.
 * @protected
 */
```

But I don't see the `opt_config` used anywhere... can we check to see how was it being used before and update the jsdoc accordingly?

packages/metal-state/test/State.js
robframpton: No, it's shouldn't be asserting `child.key1`, key1 isn't being passed a value for `Child`. I just left this assertion there to ensure that the static hint configuration of `Child` doesn't affect the values of the parent it's inheriting from.

jbalsas: Ah, I see! For some reason I assumed they had defaults!

robframpton:
@jbalsas 

Here is the commit in question regarding the inaccurate jsdoc, https://github.com/Robert-Frampton/metal.js/commit/fa26e7ccf4a2ca85c22c1c3d6c8b5fbc1ca7d14a.

Looks like it was left over by mistake, that argument was the same as what is now the [initialValues_](https://github.com/Robert-Frampton/metal.js/blob/master/packages/metal-state/src/State.js#L67) property. Everything that uses it just accesses the property now rather than passing the argument around. I'll just remove it.

jbalsas (approved):

PR #291, opened 2017-11-07
Store property on static STATE object rather than constructor

robframpton:


brunobasto:
Hey @Robert-Frampton, just out of curiosity, is there an underlying issue that this aims to solve?

robframpton:
Hey @brunobasto 

Yeah, this [original fix](https://github.com/metal/metal.js/commit/9979b997dce6d072da60d6e1e01cc26a0007884d) doesn't work for IE.

I need to look into why the tests I originally added were passing on IE though, I need to update it before we merge this in.

brunobasto:
Gotcha

yuchi:
Reminder: static class properties are not correctly inherited from superclasses in older IE since it doesn't support _setting_ the _[[Prototype]]_ of a function/class.

PR #293, opened 2017-11-10
Adds lint-staged configuration

jbalsas:
Hey @Robert-Frampton, please, take this PR with care 😂 

This applies the hopefully-final guidelines. I've separated the linting bits in packages for an easier digestion. Things to have in mind:
- I might've messed up in https://github.com/eduardolundgren/eslint-config-liferay for a lerna setup, so tests are failing on CI now... 😭 
- I've left ignored some errors because fixing them seemed risky or broke tests. Those could be considered as TODOs
- There might be a bug in prettier that prevents it from properly formatting lines of around 81-82 chars. I've ignored them for now as well
- Based on the guidelines, optional params should always have a default value (rather than the `opt_` prefix). I've left this for the future as well since it requires quite some time to go through it.


robframpton (approved):

robframpton:
Hey @jbalsas 

This looks good to me. Just waiting to merge since a high priority issue came up for the wedeploy team today when they updated to v2.14.0, I think we should probably merge and publish https://github.com/metal/metal.js/pull/298 first as a patch and then merge these changes.

jbalsas:
Hey @Robert-Frampton! We should be able to merge to `develop` and still do patch releases for these kind of issues. That was the whole point about this new branch strategy.

Just to validate, I'm going to merge this in, and then proceed to cut a release from #298... YOLO!

PR #304, opened 2017-11-20
Prepend function on metal-dom

diegonvs:
#296

Issue #296, opened 2017-11-14
metal-dom prepend


carloslancha:
It would be useful to have the opposite method of append, to prepend content to an element.

diegonvs:
I would take this issue, to get more knowledge on metal-dom and how it works :)

diegonvs:
#306

jbalsas:
Merged into `develop` via https://github.com/metal/metal.js/pull/312!

PR #306, opened 2017-11-21
Adding prepend function on metal-dom

diegonvs:
Correcting #304 to solve issue #296

### DIFF
--- a/packages/metal-dom/src/domNamed.js
+++ b/packages/metal-dom/src/domNamed.js
@@ -287,13 +287,7 @@ function isAbleToInteractWith_(node, eventName, eventObj) {
 		return false;
 	}
 
-	const matchesSelector = [
-		'BUTTON',
-		'INPUT',
-		'SELECT',
-		'TEXTAREA',
-		'FIELDSET',
-	];
+	const matchesSelector = ['BUTTON', 'INPUT', 'SELECT', 'TEXTAREA', 'FIELDSET'];
 	if (eventName === 'click' && matchesSelector.indexOf(node.tagName) > -1) {
 		return !(node.disabled || parent(node, 'fieldset[disabled]'));
 	}
@@ -376,9 +370,7 @@ export function hasClass(element, className) {
  * @private
  */
 function hasClassWithNative_(element, className) {
-	return (
-		className.indexOf(' ') === -1 && element.classList.contains(className)
-	);
+	return className.indexOf(' ') === -1 && element.classList.contains(className);
 }
 
 /**
@@ -753,10 +745,7 @@ export function toElement(selectorOrElement) {
 	) {
 		return selectorOrElement;
 	} else if (isString(selectorOrElement)) {
-		if (
-			selectorOrElement[0] === '#' &&
-			selectorOrElement.indexOf(' ') === -1
-		) {
+		if (selectorOrElement[0] === '#' && selectorOrElement.indexOf(' ') === -1) {
 			return document.getElementById(selectorOrElement.substr(1));
 		} else {
 			return document.querySelector(selectorOrElement);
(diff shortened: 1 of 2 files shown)


diegonvs:
#296

robframpton (changes requested):
Hey @diegonvs 

Can you add a test for passing a node list? Similar to what we do for the append method https://github.com/metal/metal.js/blob/master/packages/metal-dom/test/dom.js#L201-L209.

Another thought I had is how we loop over items in this function. Let's say we have a NodeList like `<div>el1</div>, <div>el2</div>, <div>el3</div>` and we pass it to this method, the NodeList will essentially get reversed because it adds each element as the first child. Then you would get something like...

```html
<div id="parent">
    <div>el3</div>
    <div>el2</div>
    <div>el1</div>
</div>
```

I think this behavior would be confusing, so let's loop through the NodeList backwards so that we preserve order.

diegonvs:
Thanks for review @Robert-Frampton, i'll preserve order on that tests

packages/metal-dom/src/domNamed.js
```diff
@@ -521,6 +521,34 @@ export function parent(element, selector) {
...
+		parent.appendChild(child);
+	}
+
+	if (isNodeListLike(child)) {
+		const childArr = Array.prototype.slice.call(child);
+		for (let i = 0; i < childArr.length; i++) {
+			parent.appendChild(childArr[i]);
+		}
+	} else {
+		parent.insertBefore(child, parent.firstChild);
```
robframpton: When calling this on a parent that has no children, this block will run even though line 537 has already added the child.

packages/metal-dom/src/domNamed.js
```diff
@@ -521,6 +521,34 @@ export function parent(element, selector) {
...
+	}
+
+	if (!isNodeListLike(child) && !isDefAndNotNull(parent.firstChild)) {
+		parent.appendChild(child);
+	}
+
+	if (isNodeListLike(child)) {
+		const childArr = Array.prototype.slice.call(child);
+		for (let i = 0; i < childArr.length; i++) {
+			parent.appendChild(childArr[i]);
```
robframpton: If you add a node list to a parent that already has a child this won't work, this test will currently fail as an example.

```javascript
        it('should prepend a node list to parent element', function() {
            let parent = document.createElement('div');
            let child = document.createElement('div');

            dom.addClasses(child, 'child');
            dom.append(parent, child);

            let childFrag = dom.buildFragment(
                '<div class="myChild">el1</div><div class="myChild2">el2</div><div class="myChild3">el3</div>'
            );

            dom.prepend(parent, childFrag.childNodes);

            assert.strictEqual(4, parent.childNodes.length);
            assert.strictEqual('myChild', parent.childNodes[0].className);
            assert.strictEqual('myChild2', parent.childNodes[1].className);
            assert.strictEqual('myChild3', parent.childNodes[2].className);
            assert.strictEqual('child', parent.childNodes[3].className);
        });
```

robframpton (changes requested):

packages/metal-dom/src/domNamed.js
diegonvs: I've corrected this using insertBefore instead appendChild

packages/metal-dom/src/domNamed.js
diegonvs: Hey @Robert-Frampton , what do you think that function return parent? User probably can get info about childNodes easily, instead actually returning child. I've sent a commit that fixes the problem on running though and returns parent.

robframpton: Hey @diegonvs, we should probably keep it the same as the `dom.append` method which returns the child.

robframpton:
Thanks @diegonvs! I've sent a follow up PR with some small formatting changes at https://github.com/metal/metal.js/pull/312

PR #307, opened 2017-11-21
Run test:saucelabs script on CI

robframpton:


jbalsas (approved):
Hey @Robert-Frampton, I'm going to merge since this is a step in the right direction :D 

I see we're still using gulp to run `metal-tools-soy`. Are you planning on getting rid of that as well and rely on npm scripts only on a different PR?

robframpton:
Hey @jbalsas 

I left it because the gulp portion isn't just compiling the soy, but it's also [modifying some paths](https://github.com/metal/metal.js/blob/develop/gulpfile.js#L11-L12). Gulp is not a bad tool for doing things like this, so I didn't see the need in changing it at this time.

PR #309, opened 2017-11-21
Default STATE values should be preserved when rendering web components. Fixes #305

### LINKED ISSUE #305: Default state values are not set by using WebComponents
## Expected behavior

The expected when using `Config.value` is use the default value when no value was passed. Using WebComponents.
```javascript
value: Config.string().value('primary');
```

Result:
```javascript
this.value;
// primary
```

## Current behavior

When no value was passed the value is `null`.
```javascript
this.value;
// null
```

robframpton:


robframpton:
The formatting commit from https://github.com/metal/metal.js/pull/307 should resolve this test failure.

packages/metal-web-component/src/define_web_component.js
```diff
@@ -80,9 +80,13 @@ export function defineWebComponent(tagName, Ctor) {
 
 			let opts = {};
 			for (let i = 0, l = observedAttributes.length; i < l; i++) {
-				opts[observedAttributes[i]] = this.deserializeValue_(
+				let deserializedValue = this.deserializeValue_(
```
jbalsas: Should this be `const`?

jbalsas (changes requested):
Hey @Robert-Frampton, not sure we ever settled for a `let` vs `const` vs `var` debate, so I'm fine either way, just asking 🤷‍♂️

PR #317, opened 2017-11-30
domData get and set

diegonvs:


packages/metal-dom/src/domData.js
```diff
@@ -10,20 +10,12 @@ class domData {
 	 * Gets Metal.js's data for the given element.
 	 * @param {!Element} element
 	 * @param {string=} name Optional property from the data to be returned.
-	 * @param {*=} initialValue Optional value to the set the requested property
-	 *     to if it doesn't exist yet in the data.
 	 * @return {!Object}
 	 */
-	static get(element, name, initialValue) {
```
robframpton: Hey @diegonvs,

I don't think we should remove any functionality from the `get` method, since that will be a breaking change (even if we account for our own use cases it might break somewhere else).

Can we maintain the `initialValue` argument, and maybe just call the new `set` method from inside `get` if the arg is passed?

packages/metal-dom/src/domData.js
```diff
@@ -35,6 +27,27 @@ class domData {
...
+	 * @param {string=} name Optional Property from the data to be returned.
+	 * @param {*=} value Optional value to the set the requested property
+	 *     to if it doesn't exist yet in the data.
+	 * @return {!Object}
+	 */
+	static set(element, name, value) {
+		if (!element[METAL_DATA]) {
+			element[METAL_DATA] = {};
+		}
+		if (!name || !value) {
```
robframpton, line 51: Wouldn't this check prevent someone from passing a falsy value?

```
domData.set(element, 'foo', true)
// foo === true
domData.set(element, 'foo', false)
// foo === true
```

robframpton (changes requested):

packages/metal-dom/src/domData.js
diegonvs: good idea Rob!

packages/metal-dom/src/domData.js
diegonvs: I'll correct this!

packages/metal-dom/src/domData.js
diegonvs: Good point @Robert-Frampton , i think in the `get` method there could be no creation or editing of domData but thinking so as not to break API, this makes sense.

diegonvs:
What do you guys think about the [`has`](https://github.com/metal/metal.js/blob/develop/packages/metal-dom/src/domData.js#L35) method?
Key value may be `0` or `""` and in this implementation will return false, i think that is broken, can i fix it?

robframpton:
Hey @diegonvs, I think the `has` method only checks to see if the `METAL_DATA` object exists. So the values shouldn't have any affect on the return value.

```javascript
return !!element[METAL_DATA];
```

packages/metal-dom/src/domData.js
robframpton: Since we're leaving the API alone for `get`, we should also leave all the instances of it alone in metal. For example, https://github.com/metal/metal.js/pull/317/commits/41756bc8e4d8ca01b99058bbcbaa770125fe4ed5 should just be left using `get`.

robframpton:
Hey @diegonvs,

At some point today we'll be cutting a release. Not trying to rush you, but what's your estimate for finishing this PR?

robframpton:
Closing in favor of #319

PR #326, opened 2017-12-05
[WIP] Reduces lifecycle for environments with no dom

### LINKED ISSUE #318: Reduce lifecycle for Server Side Rendering
Currently, the whole lifecycle is triggered on server side rendering causing a bunch of unexpected behaviors and sometimes requiring a lot of validations verifying if it's a client rendering.

For instance, React only triggers `componentWillMount` on SSR (https://reactjs.org/docs/react-component.html#componentwillmount)

If a sub-component tries to change something on `rendered` and `attached` it would cause unexpected behaviors.

The function `getCurrentData` also gets a call on SSR, and looks like it should not.
https://github.com/metal/metal.js/blob/master/packages/metal-incremental-dom/src/render/render.js#L114

On SSR `comp.element` doesn't exist, and the current code assumes it exists on that conditional.

I suggest having the same behavior, just only lifecycle to be triggered before rendering. Maybe `willAttach`.

pragmaticivan:
closes #318

### DIFF
--- a/packages/metal-component/src/Component.js
+++ b/packages/metal-component/src/Component.js
@@ -8,6 +8,7 @@ import {
 	isElement,
 	isObject,
 	isServerSide,
+	isDom,
 	isString,
 	object,
 } from 'metal';
@@ -159,8 +160,12 @@ class Component extends EventEmitter {
 		this.on('eventsChanged', this.onEventsChanged_);
 		this.addListenersFromObj_(this.dataManager_.get(this, 'events'));
 
-		this.created();
+		if (isDom()) {
+			this.created();
+		}
+
 		this.componentCreated_ = true;
+
 		if (parentElement !== false) {
 			this.renderComponent(parentElement);
 		}
@@ -211,6 +216,9 @@ class Component extends EventEmitter {
 				parent: parentElement,
 				sibling: siblingElement,
 			};
+			if (!isDom()) {
+				return this;
+			}
 			this.emit('attached', this.attachData_);
 			this.attached();
 		}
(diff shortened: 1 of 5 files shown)


robframpton:
Hey @pragmaticivan,

Maybe to reduce the amount of calls to `isDom`, we can call it once in the `Component` constructor and assign it to an instance property? Maybe call it `domExists` or something like that.

pragmaticivan:
hey @Robert-Frampton just sent some changes applying as you've suggested

pragmaticivan:
Changing that to `WIP` for more discussion about having `disposed/dispose` also `willAttach` maybe seems to be a weird name for something that will not attach anything on SSR since there's no Dom.

robframpton:
That's a good question. I wonder why React does it that way.

It seems that there might be just as much of a chance of things breaking by firing `willAttach` on the server. I wouldn't be surprised if a component did something DOM related during that lifecycle step.

pragmaticivan:
Was also thinking about leaving `created` since it represents the event in the constructor.


Wonder if you guys have any feedback on that Topic: cc @mthadley @bryceosterhaus

mthadley:
@pragmaticivan I think that it's generally a good idea to do any DOM related work inside of `attached` and `detatched`, instead of `created`. But I think you will find that sometimes developers will end up putting that kind of logic in `created` as well, even if they shouldn't.

pragmaticivan:
I agree, but wonder if that's not a good time to create a "Best Practice Guide" for developing metal components where it should be performed on SSR and client, and maybe assume that they won't use `created` for DOM related stuff.

eduardolundgren:
Even though I like the concept of simplifying server-side rendering by avoiding manual checks by the developer, lifecycle changes can mess things we can't predict. If you guys decide in moving forward discuss carefully the effects.

jbalsas:
I actually find this PR quite confusing. If I'm reading it correctly, the component lifecycle methods are being executed, but we won't be dispatching the events or will be exiting those methods quick. 

If we wanted to do this, I think it should be implemented somewhere else, so the methods on `Component` are not invoked at all. That would allow us to properly document SSR lifecycle. At the same time, if done this way, the risks @eduardolundgren is mentioning are better understood since we'd see calls disappear (risky). Just adding flow logic to the event emitters like in this PR seems quite harmless, which may not be the case.

pragmaticivan:
All good, I will close this PR and will do some research. Thanks for the feedback o/

PR #355, opened 2018-01-29
Validation for Fragment should only occur if a value exists

bryceosterhaus:


packages/metal-jsx/test/Fragment.js
```diff
@@ -3,7 +3,7 @@
 import Fragment from '../src/Fragment';
 import JSXComponent from '../src/JSXComponent';
 
-describe('Fragment', function() {
+describe.only('Fragment', function() {
```
robframpton: Excuse me sir, I think you dropped something.

bryceosterhaus:
@Robert-Frampton updated

robframpton (approved):

PR #356, opened 2018-02-07
Fix the accessibility issue when using iframe

matuzalemsteles:
hey guys, this only fixes an accessibility problem when using **iframe**.

Problem reported on https://github.com/liferay/senna.js/issues/250


julien (commented):
Looks good to me

julien (approved):

robframpton (approved):

PR #357, opened 2018-02-08
Do not favor undefined over default values when updating data manager

robframpton:


robframpton:
Note: I left the JSX test as well as the one I wrote for `ComponentDataManager`, as the `JSXDataManager` overrides the `replaceNonInternal` method that contains the fix.

jbalsas (approved):

Issue #364, opened 2018-03-09
Add the warning that the `domData.get` functionality with the `set` functionality will be deprecated in the major release


matuzalemsteles:
Hey guys,

Since `domData.set` has been added and the `domData.get` functionality has been set to `set` so that we do not have a breaking change, it would be important to add warnings that the next major release will depreciate this feature.

Something like:
```javascript
static get(element, name, initialValue) {
	if (!element[METAL_DATA]) {
		element[METAL_DATA] = {};
	}
	if (!name) {
		return element[METAL_DATA];
	}
	if (!isDef(element[METAL_DATA][name]) && isDef(initialValue)) 
		console.warn('domData: get with the set functionality will be deprecated and removed in the next major releases.');
		element[METAL_DATA][name] = initialValue;
	}
	return element[METAL_DATA][name];
}
```

What do you think?

diegonvs:
I think that we can use https://github.com/gajus/babel-plugin-log-deprecated WDYT?

diegonvs:
I think that we can try use this transform https://www.npmjs.com/package/babel-plugin-log-deprecated, what do you think?

matuzalemsteles:
Hey @diegonvs how we talk it off, being a library no longer maintained maybe not ideal and covers few use cases. In this case it's just showing a warning when you use the `set` functionality inside` get`.

I do not think we need to try to automate this, just show a friendly message to the developer 🙂.

PR #368, opened 2018-03-26
Removing IE9 and IE10 support

diegonvs:


jbalsas:
Great, @diegonvs!

We also need to go through code removals like https://github.com/metal/metal.js/blob/master/packages/metal-dom/src/domNamed.js#L413-L453

If it is supported in IE11 and the rest of our support matrix, then we can get rid of some of those fallbacks (in case there are more)

packages/metal-dom/src/features.js
```diff
@@ -56,22 +55,6 @@ class features {
 		}
 		return `${type}end`;
 	}
-
-	/**
-	 * Some browsers (like IE9) change the order of element attributes, when html
```
jbalsas: The comment says IE9, but we should check in the real browsers to see which ones actually do this.

packages/metal-dom/src/domNamed.js
```diff
@@ -205,21 +205,6 @@ export function buildFragment(htmlString) {
...
 }
 
-/**
- * Checks if the first element contains the second one.
- * @param {!Element} element1
- * @param {!Element} element2
- * @return {boolean}
- */
-export function contains(element1, element2) {
-	if (isDocument(element1)) {
```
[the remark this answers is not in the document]
matuzalemsteles: Hey @jbalsas, It would be valid to start issuing `warn` messages talking about deprecated. When the developer is using this feature. I think will be much more aware when you see this on the console when are developing your applications.

Something like proposed in #364.

packages/metal-dom/src/domNamed.js
```diff
@@ -209,15 +209,11 @@ export function buildFragment(htmlString) {
  * Checks if the first element contains the second one.
  * @param {!Element} element1
  * @param {!Element} element2
+ * @deprecated Use element1.contains(element2) directly instead of this method
```
jbalsas: Can you move this on top of the other `@param` lines? 🙏

jbalsas:
Hey @diegonvs, I just added a small comment and also, I think we should keep the `nextTick` polyfill as it is, just because it's not our code. Could you undo that?

Would this be all of the polyfills/special code for IE9-IE10 we can get rid of?

packages/metal-dom/src/domNamed.js
diegonvs: Done! :)

diegonvs:
Hey @jbalsas, i've undo deletion of polyfills for async right now.

These changes I made were the only ones I thought we could remove without compromising support for IE11, probably with the withdrawal of IE11 support in the future we will make big changes and remove many polyfills.

diegonvs:
@jbalsas, I can add in each JSDOC an annotation that represents that this function is being used as polyfill for IE11, what do you think?

jbalsas (approved):

PR #372, opened 2018-04-24
Move the `a` tag from the block to inline in HTMLParser | Fixes #371

### LINKED ISSUE #371: Some use cases do not work well when HTML2IncDom is called
### Disclaimer
When we have an application in soy that the value of our **STATE** is set to `isHTML` for `true`:
```javascript
MyComponent.STATE = {
    text: {
        isHTML: true,
        value: '<p><strong><em><u><a href="http://t4gdgreg" target="">rtetettrttrrtert</a></u></em></strong></p>'
    }
}
```

It calls the function `HTML2IncDom.buildFn(...)` to do the parser, but we have some use cases like the one above that it returns the wrong parser, this is reflected in the DOM, it seems to be some problem with the tag `a` inside the parser, but not sure about that.

This is quite common to happen in editors like AlloyEditor, to stylize the text.

### Current behavior
screen shot 2018-04-24 at 14 34 10

### Expected behavior
screen shot 2018-04-24 at 14 50 14

matuzalemsteles:
This fixes #371, with references.

* https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements
* https://www.w3schools.com/html/html_blocks.asp

### DIFF
--- a/packages/metal-incremental-dom/src/html/HTMLParser.js
+++ b/packages/metal-incremental-dom/src/html/HTMLParser.js
@@ -97,12 +97,12 @@ var empty = makeMap(
 
 // Block Elements - HTML 5
 var block = makeMap(
-	'a,address,article,applet,aside,audio,blockquote,button,canvas,center,dd,del,dir,div,dl,dt,fieldset,figcaption,figure,footer,form,frameset,h1,h2,h3,h4,h5,h6,header,hgroup,hr,iframe,ins,isindex,li,map,menu,noframes,noscript,object,ol,output,p,pre,section,script,table,tbody,td,tfoot,th,thead,tr,ul,video'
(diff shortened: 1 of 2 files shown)


brunobasto (commented):
LGTM, but wondering if we can run any kind of tests with this change applied and see if anything in Metal breaks.

yuchi:
Please pay attention that `a` is a special element that can contain both block and inline elements.

I didn't look at the source but looks like you are inferring the child type of an element based on what kind of element it is, while the spec is a little bit more complex AFAIK.

matuzalemsteles:
@brunobasto, I'm adding some cases to the tests that it broke before when it was in the `Block elements` list and adding some cases of using it as `block` and `inline`.

hey @yuchi, thanks for your concerns, as the current parser it follows this line of inferring the elements in `block` and `inline` ... I am just adding one more exception to `a` element since we can do it from `block` or `inline`. Thanks!

PR #373, opened 2018-04-25
Fixes attribute deserialization when using WebComponents | Fixes #292

### LINKED ISSUE #292: When you expect the parameter to receive `number` or `boolean`, confusions occur in WebComponents
Hey guys,

I'm joking a bit with `WebComponents` in Clay Components and I've been experiencing some difficulties in passing the parameters in some use cases at https://github.com/metal/metal-clay-components/pull/139.

#### `number`
When expected is a `number` and I try to pass in single or double quotation marks simply returns the following message:
```javascript
Warning: Error: Warning: Invalid state passed to 'minValue'. Expected type 'number', but received type 'string'. Passed to 'ClayProgressBar'. 
```

This is stranger when in the component itself I pass a `number` to `value="40"` is passed as `number` and no error is returned. 😧
#### `string`
But when expected is a `string` and I want to pass a number as `string` is returned error.
The way I found to mock it, was by passing single quotes inside double quotes. for example `label = '"10"'`.
```javascript
Warning: Error: Warning: Invalid state passed to 'label'. Expected type 'string', but received type 'number'. Passed to 'ClayBadge'.
```

#### `boolean`
When expected is `boolean` passed to paramerts `visible='false'` is return:
```javascript
Warning: Error: Warning: Invalid state passed to 'visible'. Expected type 'boolean', but received type 'string'. Passed to 'ClayModal'. 
```

### Expected behavior
Expected is matched to the cases above.

#### `number`
Pass the values as `number`.
```
minValue='10'
```

#### `string`
When I pass a `10` value that is passed as a string. How could we distinguish this?

#### `boolean`
When passing `false` value, it is like` boolean`.

matuzalemsteles:
I've been looking at issue #292 and I've been noticing that the big problem of using `boolean` when it is` false`. In return of the function **deserializeValue_** we have **OR** and as `retVal` gets **false** the return is `value` parameter that contains `'false'`.

### Note
There are some use cases that want to pass a `number` value to `string`, we can force it into the element attribute by passing `"'10'"`, but the deserializeValue_ function receives `''10''` with two single quotes and breaks the `JSON.parse`, but the correct way would be to pass `'"10"'` to do the parser correctly.

Please feel free to offer a solution to avoid this.

### Current behavior
screen shot 2018-04-25 at 18 28 29

### Expected behavior
screen shot 2018-04-25 at 18 28 59

### DIFF
--- a/packages/metal-web-component/src/define_web_component.js
+++ b/packages/metal-web-component/src/define_web_component.js
@@ -103,13 +103,11 @@ export function defineWebComponent(tagName, Ctor) {
 		 * @return {Object}
 		 */
 		deserializeValue_: function(value) {
-			let retVal;
-
 			try {
-				retVal = JSON.parse(value);
+				value = JSON.parse(value);
 			} catch (e) {}
 
-			return retVal || value;
+			return value;
 		},
 
 		/**
--- a/packages/metal-web-component/test/define_web_component.js
+++ b/packages/metal-web-component/test/define_web_component.js
@@ -1,6 +1,7 @@
 'use strict';
 
 import Component from 'metal-component';
+import core from 'metal';
 import Soy from 'metal-soy';
 import UA from 'metal-useragent';
 import {JSXComponent} from 'metal-jsx';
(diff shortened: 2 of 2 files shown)


diegonvs:
depends on #377

yuchi:
@matuzalemsteles what’s that repl-ish thing you used to make those screens?

matuzalemsteles:
hey @yuchi, I was using [Scratches](https://github.com/0x00A/scratches).

PR #384, opened 2018-06-19
Escapes html entities inside element attributes

jbalsas:
Hey @eduardolundgren, @brunobasto, what do you think about this? We aren't properly unescaping attribute values before passing it down to incremental-dom, which is causing some complex issues.

I think this should be safe since all tests are passing and I added an additional one... 🤔 

/cc @topolik


brunobasto:
LGTM, I'm curious, what kind of issues did you get before adding this? XSS?

jbalsas:
This was actually causing https://issues.liferay.com/browse/LPS-82527, because we render html, into a js string and back to html, and then we were double-escaping the actionURL of the form

brunobasto:
Ah. Got it

PR #387, opened 2018-07-24
Add script to execute styles in order | Fixes #383

### LINKED ISSUE #383: [metal-dom] Styles are not loaded in order in IE11
When using globalEvalStyles.runStylesInElement to load several CSS files and inline styles in IE11 they are not loaded in order. If you are loading a very large CSS file, and just after that some inline styles, these ones would be overriden. This is happening only in IE11.

This is happening only for styles and not for scripts, because in scripts we load them orderly, where [next file is called from previous file's callback](https://github.com/metal/metal.js/blob/master/packages/metal-dom/src/globalEval.js#L92-L127).

While for styles they are loaded inside a for loop without waiting till previous styles are completed loaded, as you can see [here](https://github.com/metal/metal.js/blob/master/packages/metal-dom/src/globalEvalStyles.js#L87-L104).

In order to find a solution, I tried a similar code in globEvalStyles but 'load' event was not being fired by IE11 for link elements, so the callback to load next styles' file was not loaded.

Thanks,

diegonvs:


diegonvs:
#383

jbalsas:
Hey guys! have you tried this in IE11? Based on our tests, this might fail to fire the proper load events there...

At least that’s what we observed when applied to senna on IE11. Could you check that?

@antonio-ortega, could you share here the portal LPS we were trying to fix with this?

antonio-ortega:
Hi @diegonvs,

I've tried your solution and it is not working. This is the first idea that came up to my mind when I started working on this and the problem with this case is we get stuck in the first iteration because apparently [load event](https://github.com/diegonvs/metal.js/blob/84dd7de188d2755b15003d9bbdefba2987b1a648/packages/metal-dom/src/globalEvalStyles.js#L66) is not fired so we don't continue looping.

The Portal LPS is[ LPS-83765](https://issues.liferay.com/browse/LPS-83765).

Thanks,

matuzalemsteles:
hey guys, Diego is seeing some other possibilities to listen when a new style was loaded on the page, it should bring more information to afternoon about some possibilities that we can go.

diegonvs:
Some updates from yesterday: Now i'm able to reproduce this and in last days i'm trying to write a solution taking as base that `styleElement.sheet.cssRules` is only populated when the file is completely loaded to avoid listening `load` event.

PR #390, opened 2018-08-08
Fix community slack invite link

diegonvs:


jbalsas:
Merging straight to `master` and updating `develop` so we can see the link right away!

PR #438, opened 2023-09-14
LPS-196120

holatuwol:
I checked various answers and https://stackoverflow.com/a/70362126 seemed like the one that made the most sense, since it could still be tested in core.js and I agreed with their assertion that toString was the least likely thing that someone would bother to implement when adding a process object to the window object for whatever code they were writing.


holatuwol:
Closing, as we'll be working around it on the portal side.

